Issue #53, opened 2026-07-27
The generated dashboard misreports: stale timestamp, invented pending decision, unknown state beside a recorded PR


atqamz:
Four defects observed in a single live `data/dashboard.md` during an ordinary two-task session. They are grouped because they share a root cause: the dashboard is a cached rendering updated as a side effect of other commands, so nothing re-derives it when the underlying `state/<id>.json` moves on.

## 1. Stale content under an authoritative timestamp

The file was roughly 30 minutes behind the task state while presenting an `Updated:` header that read as current. A stale file that looks stale is a nuisance; one that asserts freshness is a trap, and it is what led to a task being reported to the operator as awaiting action when it had already landed.

## 2. Invented Pending Decision

**Pending Decisions** carried an entry reading `stopped, reason unknown (idle 27m)`. No worker reported that. It was inferred from pane idleness and then persisted, and nothing later re-checked it.

Meanwhile the real `needs-decision` a worker had appended to its status file never appeared in that section at all. So the one section whose entire job is surfacing blocked work showed a fabricated entry and omitted the genuine one.

## 3. State `unknown` beside a PR URL and a reported `done`

One row showed state `unknown` while the same row carried a PR URL and the task's status file ended in `done:`. Three facts about one task, two of them correct, rendered side by side without anything reconciling them.

## 4. Raw decision text dumped into Recent Events

A worker's ~200-word `needs-decision` line was written into **Recent Events** verbatim as a single bullet. Recent Events is a scannable log; one entry consuming the whole section defeats it. Truncate with a pointer to the status file.

## Root cause

Defects 1 to 3 are caching artifacts. The dashboard is written by `spawn`, `pr`, `merge`, `promote`, `teardown`, `project`, and the watcher, each updating the slice it knows about. No command owns making the whole file consistent with `state/`, so any path that updates task state without a dashboard write leaves a row behind, and inferred content once written is never revisited.

`cmd/status.go` does not have these problems: it renders from `state/<id>.json` on every invocation and never reads the dashboard. That is the contrast worth designing against.

## Suggested direction, not a prescription

Rewrite the whole file from `state/` rather than patching rows, on any command that touches it. If that is too expensive on the hot path, the alternative is for the dashboard to stop carrying inferred fields entirely and render only what is literally in `state/`, with `unknown` meaning "not recorded" rather than "guessed wrong".

## Context

[...]

PR #56, opened 2026-07-27
feat(cmd): surface PR merge state in hand status

### LINKED ISSUE #51: PR merge state is observed and persisted but never surfaced in hand status
`hand watch` detects a merge correctly. `internal/watcher/events.go` `ClassifyPRMerged` fires a `pr-merged` event the first time it observes merged true, and `PersistedPRMerged` keeps it from re-firing. Confirmed live: a PR merged by hand in the GitHub UI produced `pr-merged secondhand-th-floor` in the watch stream without any prompting.

The gap is that this is the only place the fact ever appears.

`hand status <id>` for that same task, after the merge, showed:

```
State:      done
PR:         https://github.com/atqamz/secondhand/pull/49
Reported:   done: PR https://github.com/atqamz/secondhand/pull/49
```

Nothing about the merge. So the merged fact lives only in an ephemeral event stream. If nobody is reading `watch` at the moment it fires, or the supervising agent is polling `status` instead, the information is gone, and the agent goes on reporting a merged PR as awaiting merge. That happened in this session.

The state is already persisted for de-duplication purposes, so this is about surfacing rather than tracking:

- Show merge state in `hand status <id>` detail, and in the `pr` column or a state marker in the `hand status` overview.
- Ditto the dashboard's active-task table, which today shows a PR URL with no indication of whether it landed.

Related: #50 makes the same argument about events generally. A bounded, rolling event view is the wrong home for facts that are still true after they scroll off.

### LINKED ISSUE #58: Rename state.Task's Merged and PRMergedObserved to say what they mean
## Problem

`state.Task` names two related fields inconsistently, and the inconsistency is not meaningful:

```go
Merged   bool   `json:"merged"`
MergedAt string `json:"merged_at"`
PRMergedObserved bool `json:"pr_merged_observed"`
```

One carries a `PR` prefix, the other does not. Neither name states the fact it records:

- `Merged` does not mean the PR is merged. It means **hand itself performed the merge**, set at `cmd/merge.go:107` after `gh pr merge` returns success.
- `PRMergedObserved` is not the observational counterpart. Per its own comment at `internal/state/types.go:34-36` it records that **`hand watch` already announced** this PR merged, so a restarted watcher does not re-announce. `internal/watcher/watcher.go:373` latches it monotonically with `||`.

The names invite reading them as one fact plus its observation. `PRMergedObserved` can be true while `Merged` is false - that is the external-merge case, and it is the common one - so any reader who branches on `!t.Merged` to mean "not merged" is wrong for a PR that is very much merged.

The same trap caught the derived display type on #56 three times: once on the enum value names, once on the parameter names, once here. Each was fixed there by naming the facts rather than mirroring the fields. The source is still unfixed.

## Proposal

**Both fields record actions hand took, not facts about the PR.** One says hand ran the merge; the other says `hand watch` emitted an announcement. Neither is a claim about the PR's state on GitHub. Naming them that way makes the symmetry truthful rather than merely tidy, and leaves no implication to misread.

The precedent is one field below, at `internal/state/types.go:43`: `DoneVerified` records that `hand watch` announced the verified done line. Hand's own bookkeeping, in `<thing><pastParticiple>` form.

```go
MergeExecuted   bool   `json:"merged"`
MergeExecutedAt string `json:"merged_at"`
MergeAnnounced  bool   `json:"pr_merged_observed"`
```

**JSON tags stay exactly as they are.** Go identifier rename only, so no schema change and no migration of existing `state/*.json`.

### Rejected names

- `MergedByHand` / `MergedByHandAt` - "by hand" idiomatically means *manually, by a human*, the exact opposite of what the field records.
- `prMerged` / `prMergedObserved` - implies the pair differs only by observation when it differs by actor.
- `Merged` / `MergedAnnounced` - a genuine improvement over the status quo, and the smaller diff since `cmd/merge.go` goes untouched. It fixes the stray `PR` prefix and the wrong "Observed", but leaves `Merged` naming a provenance fact as though it were a status fact, which is the larger half of the problem. Live alternative if the narrower diff is preferred.

## Why this is its own PR

The rename touches `internal/state/types.go`, `cmd/merge.go`, `internal/watcher/watcher.go`, `cmd/status.go`, `internal/dashboard`, and their tests. Folding it into #51's surfacing work would have made that diff unreviewable, and a mechanical rename mixed with behaviour changes hides both.

`Related: #51`

## Definition of done

- Identifiers renamed across all readers and writers; no call site left on the old names.
- Every JSON tag byte-identical to before, verified by round-tripping an existing `state/*.json` and diffing.
- The field comments explain the actor distinction without restating the identifier.
- Existing tests pass unchanged in behaviour; assertions updated only where they name an identifier.

atqamz:
## Summary

- `hand status` now shows PR merge state directly, in both the fleet overview and single-task
  detail: `state.Task`'s `MergeExecuted` (hand ran the merge) and `MergeAnnounced` (hand watch's
  own `gh` poll saw someone else merge it) render as a distinguishable, non-accusatory suffix on
  the PR column - ` (merged)` or ` (merged, external)`. It reads `state.Task` directly on every
  invocation, so this closes #51 on its own, with no dashboard change.
- Folds in #58: renames `state.Task`'s `Merged`/`MergedAt`/`PRMergedObserved` to
  `MergeExecuted`/`MergeExecutedAt`/`MergeAnnounced`. JSON tags are byte-identical (`merged`,
  `merged_at`, `pr_merged_observed`) - Go identifier rename only, no schema change, no migration.

## Why the dashboard half is gone

This PR originally also wired the merge marker into `internal/dashboard`'s `Render`/`Parse` round
trip, `hand merge`'s post-merge reconcile, and `hand watch`'s `pr-merged` handling. Dropped
entirely: #62 deletes `data/dashboard.md` and `internal/dashboard` outright rather than fixing the
read-modify-write round trip that this PR's own gate review had already found losing this exact
flag when a row's PR cell went stale (`marker-dropped-when-pr-cell-empty` - fixed at the time, in
code that would have been deleted within days of landing). `internal/dashboard` now carries none
of this PR's changes: reverted `dashboard.go`, `dashboard_test.go`, `cmd/merge.go`,
`cmd/merge_test.go`, `internal/watcher/watcher.go`, `watcher_test.go` to `main`, and decoupled
`cmd/status.go`'s `mergeSuffix` from `dashboard.DeriveMergeState` so it derives the suffix from
`state.Task` directly.

The design question this raised - whether `Merged`/`PRMergedObserved` round-tripping through
rendered markdown was the right shape at all - was posted to #53 as a question rather than decided
there: [comment](https://github.com/atqamz/secondhand/issues/53#issuecomment-5091646707). #62's
decision to delete the file rather than patch the round trip is the answer.

## Wording

- No PR, or PR unmerged: unchanged, no suffix.
- `MergeExecuted`: ` (merged)`.
- `MergeAnnounced` alone: ` (merged, external)`.
- Both set: renders as ` (merged)`. Once hand performed the merge, a later poll independently
  confirming the same merge is bookkeeping, not a new fact - it doesn't downgrade who did it.

`MergeAnnounced` alone overwhelmingly means the operator merged the PR in the GitHub web UI - a
normal, common path, not an incident. "external" was chosen over anything that could read as an
alarm, since the wording must not imply wrongdoing (per #51). It still reads unambiguously next to
a PR link: the merge happened, and `hand merge` did not do it - which is exactly the fact a
supervising agent needs to decide whether to look closer.

## The rename (#58)

`state.Task` named two related facts inconsistently, and neither name stated the fact it
recorded: `Merged` did not mean the PR is merged, it meant hand itself performed the merge;
`PRMergedObserved` was not that fact's observational counterpart, it meant hand watch already
announced the merge (its own idempotence latch, so a restarted watcher doesn't re-announce). Both
fields record actions hand took, not facts about the PR, so they're renamed to say that:
`MergeExecuted`/`MergeExecutedAt`/`MergeAnnounced`. `DoneVerified` is the existing precedent for
this naming shape.

The same trap the rename fixes had already caught the derived display type on this PR three times
- once on enum value names, once on derivation parameter names, once on a field comment - before
the source fields themselves were touched. Renaming after landing new code against names already
known to be wrong is the worse order, and with the dashboard half gone the rename surface is small
enough (`internal/state/types.go`, `cmd/merge.go`, `internal/watcher/events.go`,
`internal/watcher/watcher.go`, `cmd/status.go`, and their tests - about fifteen identifier sites)
to fold in here rather than ship separately, which is the fold #58 itself now proposes.

JSON tags stay exactly as they were (`merged`, `merged_at`, `pr_merged_observed`): this is a Go
identifier rename only. It requires no migration of existing `state/*.json` - Go field names and
JSON tags are independent, and this PR is the proof: no file on disk needs to change for any of
this to work.

## Before / after

`hand status` overview, same three tasks throughout: `task-hand-merged` (`MergeExecuted=true`),
`task-merged-externally` (`MergeAnnounced=true`, `MergeExecuted=false` - the operator merged it in
the GitHub web UI), `task-awaiting-merge` (neither set).

Before:
```
task-awaiting-merge      nsr  ship  unknown  2h ago
task-hand-merged         nsr  ship  unknown  2h ago
task-merged-externally   nsr  ship  unknown  2h ago
```

After:
```
task-awaiting-merge      nsr  ship  unknown                     2h ago
task-hand-merged         nsr  ship  unknown (merged)            2h ago
task-merged-externally   nsr  ship  unknown (merged, external)  2h ago
```

`hand status <id>` detail carries the same suffix on the `PR:` line.

## Scope

Surfacing only, as scoped by #51. `internal/watcher`'s classification (`ClassifyPRMerged`,
`PersistedPRMerged`, `dashboardRowFor`'s `KindPRMerged` disposition) is untouched.
`internal/dashboard` carries none of this PR's changes - #62 owns that file now. No new `hand`
subcommand.

## Gate

Ran through the `no-mistakes` gate against this reshaped diff (a behaviour change: the rename
touches every reader and writer of the three fields).

Closes #51
Closes #58

## Test plan

- `cmd/status_test.go`: detail view shows the marker for a merged PR and is unchanged for an
  unmerged one; all four `MergeExecuted`/`MergeAnnounced` combinations; fleet overview renders the
  marker; a task with no PR at all is unaffected.
- `internal/state`, `internal/watcher`, `tests/e2e`: assertions updated to the renamed
  identifiers only, no behavior change.
- `make lint`, `go test -race ./...`, and `make e2e` all pass locally, and again inside the
  `no-mistakes` gate run.

### DIFF
--- a/cmd/status.go
+++ b/cmd/status.go
@@ -75,10 +75,6 @@ type statusJSON struct {
 	ReportHistory  []string      `json:"report_history,omitempty"`
 }
 
-// mergeSuffix distinguishes who merged the PR: hand itself (MergeExecuted) or
-// something else that hand watch's own gh poll saw (MergeAnnounced) - the same
-// wording that recorded a human merging in the GitHub web UI as a normal path
-// rather than an incident.
 func mergeSuffix(t state.Task) string {
 	switch {
 	case t.PR == "":
--- a/cmd/status.go
+++ b/cmd/status.go
@@ -60,17 +60,36 @@ type reportedJSON struct {
 }
 
 type statusJSON struct {
-	ID            string        `json:"id"`
-	Project       string        `json:"project"`
-	Kind          string        `json:"kind"`
-	Harness       string        `json:"harness,omitempty"`
-	AgentState    string        `json:"agent_state"`
-	Worktree      string        `json:"worktree"`
-	Herdr         state.Herdr   `json:"herdr"`
-	PR            string        `json:"pr"`
-	CreatedAt     string        `json:"created_at"`
-	Reported      *reportedJSON `json:"reported,omitempty"`
-	ReportHistory []string      `json:"report_history,omitempty"`
+	ID             string        `json:"id"`
+	Project        string        `json:"project"`
+	Kind           string        `json:"kind"`
+	Harness        string        `json:"harness,omitempty"`
+	AgentState     string        `json:"agent_state"`
+	Worktree       string        `json:"worktree"`
+	Herdr          state.Herdr   `json:"herdr"`
+	PR             string        `json:"pr"`
+	MergeExecuted  bool          `json:"merged"`
+	MergeAnnounced bool          `json:"pr_merged_observed"`
+	CreatedAt      string        `json:"created_at"`
+	Reported       *reportedJSON `json:"reported,omitempty"`
+	ReportHistory  []string      `json:"report_history,omitempty"`
+}
+
+// mergeSuffix distinguishes who merged the PR: hand itself (MergeExecuted) or
+// something else that hand watch's own gh poll saw (MergeAnnounced) - the same
+// wording that recorded a human merging in the GitHub web UI as a normal path
+// rather than an incident.
+func mergeSuffix(t state.Task) string {
+	switch {
+	case t.PR == "":
+		return ""
+	case t.MergeExecuted:
+		return " (merged)"
+	case t.MergeAnnounced:
+		return " (merged, external)"
+	default:
+		return ""
+	}
 }
 
 func runStatusFleet(cmd *cobra.Command, home string, client *herdr.Client, asJSON bool) error {
@@ -92,10 +111,11 @@ func runStatusFleet(cmd *cobra.Command, home string, client *herdr.Client, asJSO
 		rows = append(rows, statusJSON{
 			ID: t.ID, Project: t.Project, Kind: t.Kind, Harness: t.Harness,
 			AgentState: agentState,
-			Worktree:   t.Worktree, Herdr: t.Herdr, PR: t.PR, CreatedAt: t.CreatedAt,
+			Worktree:   t.Worktree, Herdr: t.Herdr, PR: t.PR,
+			MergeExecuted: t.MergeExecuted, MergeAnnounced: t.MergeAnnounced, CreatedAt: t.CreatedAt,
 			Reported: reportedFrom(last, len(lines) > 0, readErr),
 		})
-		suffixes = append(suffixes, reportSuffix(agentState, reported, reportedOK, readErr))
+		suffixes = append(suffixes, reportSuffix(agentState, reported, reportedOK, readErr)+mergeSuffix(t))
 	}
 
 	if asJSON {
@@ -179,7 +199,8 @@ func runStatusSingle(cmd *cobra.Command, home string, client *herdr.Client, id s
 	if asJSON {
 		out := statusJSON{
 			ID: t.ID, Project: t.Project, Kind: t.Kind, Harness: t.Harness,
-			AgentState: agentState, Worktree: t.Worktree, Herdr: t.Herdr, PR: t.PR, CreatedAt: t.CreatedAt,
+			AgentState: agentState, Worktree: t.Worktree, Herdr: t.Herdr, PR: t.PR,
+			MergeExecuted: t.MergeExecuted, MergeAnnounced: t.MergeAnnounced, CreatedAt: t.CreatedAt,
 			Reported: reportedFrom(last, len(tail) > 0, readErr), ReportHistory: history,
 		}
 		enc := json.NewEncoder(cmd.OutOrStdout())
@@ -190,6 +211,8 @@ func runStatusSingle(cmd *cobra.Command, home string, client *herdr.Client, id s
 	pr := t.PR
 	if pr == "" {
 		pr = "(none)"
+	} else {
+		pr += mergeSuffix(t)
 	}
 	reported := "(none)"
 	switch {
--- a/SPECS.md
+++ b/SPECS.md
@@ -540,8 +540,8 @@ Behavior (PR merge, default):
 4. Refuse if checks are not green.
 5. Run `gh pr merge <number> --repo <owner/repo> --squash` (or specified method).
 6. Update `state/<id>.json` with merge status.
-7. Update `data/dashboard.md`.
-8. Run `hand project sync <project>` to fast-forward the project clone.
+7. Run `hand project sync <project>` to fast-forward the project clone.
+8. Refresh the dashboard's Projects section, only if that sync advanced the clone.
 
 Behavior (local merge, `--local`):
 1. Read `state/<id>.json` for worktree and project.
--- a/SPECS.md
+++ b/SPECS.md
@@ -366,7 +366,8 @@ Behavior (fleet overview):
 1. List all `state/*.json` files.
 2. For each, query herdr for current agent state.
 3. If agent state is `idle` or `done` (herdr's two spellings of "pane stopped being busy" - see "Agent state" below), consult the task's last report line (see "Report channel"): no report, or the last line was still `working`, appends ` (unreported)`; any other terminal report appends ` (reported: <state>)`; a report file that exists but can't be read appends ` (report unreadable)`, never ` (unreported)` - an I/O fault is not evidence the worker never reported. Any other agent state is printed unadorned.
-4. Print one line per task.
+4. If the task has a recorded PR, append its merge state to the same column: ` (merged)` when `hand` performed the merge, ` (merged, external)` when only `hand watch`'s own `gh` poll saw it merged. It is appended whatever the agent state is, since a merged PR is a fact about the PR rather than about the pane.
+5. Print one line per task.
 
 Output (fleet overview):
 ```
@@ -375,6 +376,7 @@ dark-mode       nsr     ship    blocked     45m ago
 stuck-task      nsr     ship    idle (unreported)      1h ago
 paused-task     nsr     ship    idle (reported: needs-decision)   30m ago
 investigate     nsr     scout   done (reported: done)      10m ago
+shipped-fix     nsr     ship    done (reported: done) (merged, external)   5m ago
 ```
 
 Behavior (single task):
@@ -402,6 +404,8 @@ Report history (reported by worker, not verified current truth):
   needs-decision: two ways to fix the race, ask-user found both risky
 ```
 
+The `PR:` line reads `(none)` with no PR recorded, and otherwise carries the same merge suffix the fleet overview appends: `PR:         https://github.com/org/repo/pull/42 (merged, external)`.
+
 The "Report history" label is deliberate: these lines are the worker's own claims about itself, not something `hand` has verified, same caution as the `done`-vs-`reported-done` distinction in `hand watch`.
 
 Output (JSON, single task):
@@ -415,6 +419,8 @@ Output (JSON, single task):
   "worktree": "/home/user/.treehouse/nsr-abc/1/nsr",
   "herdr": {"session": "default", "tab_id": "wA:tB", "pane_id": "wA:pC"},
   "pr": "",
+  "merged": false,
+  "pr_merged_observed": false,
   "created_at": "2026-07-24T08:00:00Z",
   "reported": {"state": "needs-decision", "note": "two ways to fix the race, ask-user found both risky"},
   "report_history": ["working: added the retry loop", "needs-decision: two ways to fix the race, ask-user found both risky"]
@@ -544,7 +550,8 @@ Behavior (local merge, `--local`):
 4. In the project clone: `git merge --ff-only <task-branch>`.
 5. Refuse if fast-forward is not possible (diverged branches).
 6. Update `state/<id>.json` with merge status.
-7. Update `data/dashboard.md`.
+
+A local merge writes no dashboard row.
 
 Output:
 ```
@@ -847,7 +854,7 @@ Updated: 2026-07-24T12:30:00Z
 | `hand status` | Refresh agent states in Active Tasks (only when dashboard is stale) |
 | `hand send` | No update |
 | `hand teardown` | Move from Active Tasks to Recent Completions (keep last 10), drop the task's Pending Decisions entry |
-| `hand merge` | Update PR status, add to Recent Events |
+| `hand merge` | Refresh Projects when the follow-up project sync advanced the clone. `--local`: no update |
 | `hand pr` | Set PR on the task's row |
 | `hand watch` | Update agent states, add actionable events to Recent Events (keep last 20), update Pending Decisions, auto-record a PR seen on the report channel |
 | `hand project add` | Add to Projects |
--- a/cmd/merge.go
+++ b/cmd/merge.go
@@ -48,7 +48,7 @@ func newMergeCmd() *cobra.Command {
 				return asPrecondition(err)
 			}
 
-			if t.Merged {
+			if t.MergeExecuted {
 				return &ExitError{Err: fmt.Errorf("task %s already merged", t.ID), Code: 3}
 			}
 
@@ -104,8 +104,8 @@ func runPRMerge(cmd *cobra.Command, home string, t state.Task, method string) er
 		return fmt.Errorf("gh pr merge failed: %s", strings.TrimSpace(string(out)))
 	}
 
-	t.Merged = true
-	t.MergedAt = time.Now().UTC().Format(time.RFC3339)
+	t.MergeExecuted = true
+	t.MergeExecutedAt = time.Now().UTC().Format(time.RFC3339)
 	if err := state.Write(home, t); err != nil {
 		return fmt.Errorf("write task state: %w", err)
 	}
@@ -177,8 +177,8 @@ func runLocalMerge(cmd *cobra.Command, home string, t state.Task) error {
 		return &ExitError{Err: fmt.Errorf("fast-forward not possible: %s", strings.TrimSpace(string(out))), Code: 3}
 	}
 
-	t.Merged = true
-	t.MergedAt = time.Now().UTC().Format(time.RFC3339)
+	t.MergeExecuted = true
+	t.MergeExecutedAt = time.Now().UTC().Format(time.RFC3339)
 	if err := state.Write(home, t); err != nil {
 		return fmt.Errorf("write task state: %w", err)
 	}
--- a/cmd/merge_test.go
+++ b/cmd/merge_test.go
@@ -185,7 +185,7 @@ func TestMergePRRefusesRedCI(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
-	if got.Merged {
+	if got.MergeExecuted {
 		t.Fatal("want task not marked merged")
 	}
 }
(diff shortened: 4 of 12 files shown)


cmd/status.go
```diff
@@ -60,17 +60,28 @@ type reportedJSON struct {
...
+	ID               string        `json:"id"`
+	Project          string        `json:"project"`
+	Kind             string        `json:"kind"`
+	Harness          string        `json:"harness,omitempty"`
+	AgentState       string        `json:"agent_state"`
+	Worktree         string        `json:"worktree"`
+	Herdr            state.Herdr   `json:"herdr"`
+	PR               string        `json:"pr"`
+	Merged           bool          `json:"merged"`
+	PRMergedObserved bool          `json:"pr_merged_observed"`
```
atqamz: I think Merged and PRMergedObserved better be enum?

atqamz (commented):
reviwed

cmd/status.go
atqamz: Not for storage, yes for display. They look like a 2-bit encoding of one 3-state fact, but they are not.

`Merged` is set by `hand merge` at `cmd/merge.go:107` after `gh pr merge` returns success, alongside `MergedAt`. It means hand performed the merge.

`PRMergedObserved` is not the counterpart fact. Per its own comment at `internal/state/types.go:34-36`, it records that `hand watch` already announced this PR merged, and it exists so a restarted watcher does not re-announce. `internal/watcher/watcher.go:373` latches it monotonically with `||`. It is an announcement-idempotence flag, not merge provenance.

So all four combinations are reachable and meaningful, including both-true: hand merges, then a later `watch` poll sees the merged PR and latches its announcement flag. An enum would have to pick one of those two facts to keep. Collapsing them loses the watcher's dedup state, which is what stops a restart from replaying announcements, and it would need a migration on every existing `state/*.json`.

Where an enum does belong is the derived display state, and that is a real improvement over what this PR has now. `MergeMarker` returns a raw string that two packages consume, so `cmd/status.go` and `internal/dashboard` can drift and the tests assert on substrings. A small `MergeState` type with `Unmerged`, `MergedByHand`, `MergedExternal`, derived once from the two fields, gives one place where the three-way decision lives and lets tests assert a value. The storage keeps both facts; the display gets the enum.

Doing that now on this branch.

atqamz: Done in be3cb85. state.Task's Merged and PRMergedObserved are untouched - still two independent booleans, storage schema unchanged. Added dashboard.MergeState (Unmerged, MergedByHand, MergedExternal) and DeriveMergeState(merged, prMergedObserved bool) MergeState as the single place that makes the three-way call; MergeMarker and cmd/status.go's mergeSuffix both go through it now instead of each reaching their own conclusion from a raw string. Tests assert on the MergeState value for all four boolean combinations (both-true still derives to MergedByHand), with the round-trip and fleet-overview tests kept as-is to pin the actual rendered wording. Pushed and PR body updated with a Storage vs display section.

internal/dashboard/dashboard.go
```diff
@@ -46,26 +46,26 @@ type ActiveTask struct {
...
+// saw the PR merged some other way - overwhelmingly the operator merging in the
+// GitHub web UI, a normal and common path, not an incident. The wording never reads
+// as an alarm. Merged takes priority when both are set: hand watch's poll
+// independently observing a merge hand itself already performed doesn't downgrade
+// the fact of who did it.
 func MergeMarker(merged, prMergedObserved bool) string {
 	switch {
 	case merged:
 		return mergedSuffix
 	case prMergedObserved:
```
atqamz: should it merged and mergedObserved?

atqamz (commented):
reviewed

internal/dashboard/dashboard.go
atqamz: Deliberately not, and the both-true case is the reason.

Naming the display values after the storage fields would imply a correspondence that does not hold. `MergedExternal` is derived from `prMergedObserved && !merged`, not from `PRMergedObserved` being set. A PR that `hand merge` merged and that a later `watch` poll then observed has `PRMergedObserved == true` and derives to `MergedByHand`. So a value called `MergedObserved` would be read as "the observed flag is set" and would be wrong exactly where it matters most.

That is the same conflation the storage discussion above was about, arriving one layer up. The two booleans answer "did hand do it" and "has watch already announced it". The three display values answer a different question: "what should the operator read in this cell". Keeping the vocabularies distinct is what stops the next reader from assuming a field maps to a value.

`MergedByHand` and `MergedExternal` also say the operator-relevant thing directly, and `MergedExternal` matches the rendered wording `(merged, external)`, so the enum and the output agree.

One real inconsistency this did surface, in the same file. `ActiveTask`'s field comment at `internal/dashboard/dashboard.go:42-44` says `PRMergedObserved` means the gh poll "saw the PR merged some other way". The `MergeState` comment 20 lines below correctly calls it an announcement-idempotence latch. The first is the loose characterisation that makes the enum-mirroring idea look right; "some other way" is the derived state, not the field. Fixing that comment to match.

atqamz: This one is a fair hit, and neither of your two options is quite it.

The asymmetry is inherited, not invented here. `state.Task` has `Merged` and `PRMergedObserved`, so the parameter list copied the mismatch up one layer. That is the whole explanation, which means it is noise rather than signal, and you were right to stop on it.

Both symmetric options lose something, though:

- `prMerged` / `prMergedObserved` reads as "the PR is merged" and "we observed it merged". But the first does not mean the PR is merged, it means *hand performed the merge*. Symmetry here would imply the pair differs only by observation, when they differ by actor.
- `merged` / `mergedObserved` has the same problem in the other direction, and `mergedObserved` still describes the field as an observation when what it actually is, per `state.Task`'s own comment, is an announcement latch for `hand watch`.

Making two non-parallel facts look parallel is the same failure as naming the enum after the fields. What kills the confusion is naming them for what they are, so the reader never has to ask why they differ:

```go
func DeriveMergeState(handMerged, watchAnnounced bool) MergeState
```

Now the asymmetry is gone and the difference is stated. Doing that.

The wider version of your point stands and I am not folding it into this PR: `state.Task.Merged` and `PRMergedObserved` have the same problem at the source, and renaming the Go identifiers there (JSON tags unchanged, so no migration) would touch `cmd/merge.go`, `internal/watcher`, and their tests. That is a rename PR, not a surfacing one. Filing it separately so this branch stays reviewable.

atqamz: `MergedAnnounced` is better than my `MergeAnnounced` and it kills the right half of the problem. And my `MergedByHand` was a bad name: "by hand" idiomatically means *manually, by a human*, which is the exact opposite of what the field records. Withdrawn.

But keeping `Merged` leaves the defect that motivates the rename. `MergedAnnounced` can be true while `Merged` is false - that is the external-merge case, and it is the common one. So the pair still reads as though `MergedAnnounced` implies `Merged`, and someone will eventually branch on `!t.Merged` for a PR that is very much merged. That is the bug the current names already invite.

What I think actually resolves it: **both fields record actions hand took, not facts about the PR.**

- One says hand ran the merge.
- The other says hand watch emitted an announcement.

Neither is a claim about the PR's state on GitHub. Once they are named that way, the symmetry becomes truthful instead of merely tidy, and there is no implication left to misread.

This repo already has the precedent, one field below: `DoneVerified` at `internal/state/types.go:43` records that hand watch announced the verified done line. It is hand's bookkeeping in `<thing><pastParticiple>` form, not a statement about the world. So:

```go
MergeExecuted   bool   `json:"merged"`
MergeExecutedAt string `json:"merged_at"`
MergeAnnounced  bool   `json:"pr_merged_observed"`
```

Same shape as `DoneVerified`, symmetric, and each name is true on its own without reference to the other. JSON tags unchanged, so still no migration.

If you would rather keep the diff to one field, `Merged` + `MergedAnnounced` is a real improvement over what is there now and I will take it - it fixes the stray `PR` prefix and the wrong "Observed". I just do not think it is worth opening a rename PR for the smaller half of the problem.

Updating #58 with this. Your call on which shape lands.

atqamz: Done in b4642a0: DeriveMergeState(handMerged, watchAnnounced bool), MergeMarker updated to match since it's a direct pass-through with the same asymmetric params. Doc comment reworded to the new names, same length. Also pushed 3a6ff22 fixing the ActiveTask.PRMergedObserved field comment your earlier question surfaced. Not touching state.Task's identifiers here per your note - that's your follow-up issue. PR body updated.

atqamz: The confusion is mine - I kept relitigating a small point. Going with your `Merged`, `MergedAt`, `MergedAnnounced`. Straight answers to the two you asked, then the reason I dropped my objection.

**Why not `MergedInternal` / `MergedExternal`:** nothing would ever write `MergedExternal`. There is no code path that learns "someone else merged this". The external case is *inferred* from `announced && !executed` - `hand watch`'s poll only reports that the PR is merged, never by whom. A stored field nobody sets is a field that is always false.

**Why not an enum in storage:** the two facts co-occur. `hand merge` sets one, then a later `watch` poll sets the other on the same task. All four combinations are reachable:

| hand ran the merge | watch announced it | means |
|---|---|---|
| no | no | nothing known |
| yes | no | hand merged, watch has not announced yet |
| no | yes | merged by someone else - you, on the web |
| yes | yes | hand merged, watch later announced it |

An enum needs mutually exclusive states. These are not, so it would have to drop one of the two facts. That is the only reason storage stays two booleans. The enum *is* right for the three-way display fact, which is what `DeriveMergeState` now does.

**And my objection was theoretical, which I should have checked before pushing it twice.** I claimed someone would eventually read `!t.Merged` as "not merged". Every existing reader:

- `cmd/merge.go:51` - `if t.Merged` guarding against a double merge. Correct either way.
- `internal/watcher/events.go:247` - `t.Merged || (t.PR != "" && ts.PRMerged)`, which already ORs them, with a comment at 239 explaining exactly why it has to.

So nothing makes the mistake, and the one place needing "merged at all" already handles it. Your names fix the two real defects, the stray `PR` prefix and the wrong "Observed", in a one-field diff that leaves `cmd/merge.go` alone. That is the better trade and I was over-weighting a hypothetical.

#58 updated to your shape.

atqamz: Retracting the reply above - I conceded to close the thread rather than because I had finished reasoning. Finishing it now.

I went looking for whether the field is really two facts welded together, because if it were, the answer would be to split it. `internal/watcher/watcher.go:113` settles that:

```go
if t.PR != "" && !ts.PRMerged {   // poll is skipped once the latch is set
```

and `ClassifyPRMerged` at `events.go:161-167` sets the latch in the same breath as emitting the event. So observation, announcement, and stop-polling are welded by construction: you cannot observe without announcing, and once announced the poll never runs again. It is one coherent fact - **the poll saw this PR merged** - with the announcement and the stop as consequences.

That kills the three-field split I was about to propose (`Executed` / `Observed` / `Announced`). It would add a field and a migration for zero behavioural difference today, justified only by a hypothetical future quiet-mode. Not worth it.

But it also kills `MergedAnnounced`, and this is where I disagree with your name rather than defer to it. The announcement is the *consequence*; the observation is the *cause*, and the observation is what the display actually consumes. Naming the field for its side effect points the next reader at the least important of its three roles - and specifically at the one that is pure internal bookkeeping. `Observed` was never the wrong word.

The word that *is* wrong is on the other field. `Merged` does not mean the PR is merged, it means hand ran the merge. That was always the bigger of the two defects, and it is the one both our proposals kept dodging.

So:

```go
MergeExecuted   bool   `json:"merge_executed"`
MergeExecutedAt string `json:"merge_executed_at"`
MergeObserved   bool   `json:"merge_observed"`
```

Symmetric, both `Merge<PastParticiple>`, one says hand did it and the other says hand saw it. Neither implies the other. Your asymmetry complaint is answered, the stray `PR` prefix is gone, no field is added, and no speculative structure is introduced.

Tags changed too, as you said. That does make it a schema change, so migration is now in scope rather than out of it: six live `state/*.json` in this home carry `pr_merged_observed` today, and other homes exist. Read the legacy keys when the new ones are absent, write only the new ones, and files converge on the next write. `hand status --json` also emits `merged` and `pr_merged_observed` as output keys, which is an operator-visible contract - it changes with them, and that has to be called out rather than slipped in.

Rewriting #58 to this.

atqamz:
Context from the review that has not been applied to this PR yet.

The dashboard half of this change is on the way out: #62 deletes `data/dashboard.md` and `internal/dashboard` outright, rather than fixing the read-modify-write round trip that made a merge marker necessary in the first place. That means `Render`/`Parse` marker handling, both `ActiveTask` merge fields, `SetMergeState`, `MergeStateUpdate`, and `TestRenderParseRoundTripPreservesMergeFlagOnPRLessRow` are all code that would be deleted shortly after landing.

The `hand status` half is unaffected and closes #51 on its own: it reads `state.Task` directly and needs no dashboard change.

Recommendation is to split rather than merge as is. Separately, whichever half survives should carry the naming fix currently deferred to #58 - the stated blocker there, that renaming would force a migration on every `state/*.json`, is not real, since the Go field name and its JSON tag are independent.

Awaiting Atqa's call before anything is changed on the branch.

cmd/status.go
atqamz: Answered on the display side, declined on the storage side.

Display is an enum now. `mergeSuffix` in `cmd/status.go` is a single switch over one task returning one of three values - no suffix, ` (merged)`, ` (merged, external)` - so the four boolean combinations collapse to three states in exactly one place, and the both-true case is decided there rather than at each call site.

Storage stays two booleans, and the rename this PR folds in (#58) is what makes that defensible. They are not two views of one fact: `MergeExecuted` is written by `hand merge` when it runs the merge, `MergeAnnounced` is a latch written by `hand watch` when it emits the announcement. Different actors, different times, and either can be set without the other. An enum would have to encode "announced but not executed" and "executed but not yet announced" as distinct members anyway, so it buys nothing and costs a real JSON schema change plus a migration of every existing `state/*.json` - the exact cost this PR avoids by keeping the tags byte-identical.

The original names are what made the enum look right. `Merged` and `PRMergedObserved` read as one fact plus its observation. `MergeExecuted` and `MergeAnnounced` read as two actions hand took, which is what they are.

internal/dashboard/dashboard.go
atqamz: Moot. `internal/dashboard` carries none of this PR's changes any more - the dashboard half was reverted to base, since #62 deletes `data/dashboard.md` and the package outright.

The naming question itself is answered in the other thread and in the folded #58: the fields are now `MergeExecuted` and `MergeAnnounced`.

PR #82, opened 2026-07-28
feat(watcher): add hand watch --until-event, parked detection, and pane-scoped state resets

### LINKED ISSUE #75: hand watch cannot wake its orchestrator, so events are only seen if someone polls
## What happened

On 2026-07-28 four workers ran for about two hours with `hand watch` attached. The watcher was correct throughout: it emitted `reported-done nsr-mp-join-fix: https://github.com/yes2games/nsr/pull/2852`, and `idle-unreported` for every worker that went quiet without a terminal report.

None of it reached the orchestrator. Atqa found PR 2852 and PR 2854 himself and had to re-prompt the session twice to get them acted on.

## Why

`hand watch` is a long-running process. An orchestrator session running it as a background task is only re-invoked when that process **exits**. `watch` never exits, so its stdout accumulates in a file that is read only if the orchestrator independently decides to go look. There is no push. "Remember to check the watcher" is not a mechanism, and it is what failed here.

## The gap this is not

- atqamz/secondhand#70 is a worker reaching a terminal state with **no** watcher attached. Here the watcher was attached and working.
- atqamz/secondhand#73 is `hand watch` not being a singleton. Unrelated to delivery.

This is the third distinct hole in the same path, which suggests the watcher's delivery contract, not just its detection, is what needs designing.

## Current workaround, in place now

The watcher runs piped through `tee` to `state/watch.log` and then `grep -m1` on the reportable events, so the pipeline exits on the first one and the exit re-invokes the orchestrator. It works, but it wakes on one event and then must be restarted by hand, and `idle-unreported` is deliberately excluded from the trigger because it repeats every poll and would cause a wake storm. That exclusion means the exact signal that was missed on 2026-07-28 still does not wake anyone.

## What to work out

The design question is how `watch` delivers, not what it detects. Candidates, none obviously right:

- A `--until-event` mode that exits on the first reportable transition, making the exit itself the delivery. Composes with how background tasks already notify, but one event per run.
- Debounced or edge-triggered `idle-unreported` so a worker going quiet fires once, not every poll, and can safely be a wake trigger.
- An out-of-band channel that does not depend on process exit at all.

Whatever it is, the acceptance test is concrete: a worker that goes idle without a terminal report must reach the orchestrator without a human prompting it.

atqamz:
## Intent

From-scratch review of the full atqamz/secondhand#75 change (hand watch --until-event) on a head nobody hand-patched afterwards. Confirm: (1) the exit-code fix distinguishing arm-timeout/signal (exit 4) from a named per-task probe failure (exit 5) in connect/probeAllTasks, (2) status_changed_for durability tying a resumed dwell to the status it was recorded for, (3) the config-seconds dedup in cmd/watch.go, all hold correctly. Separately, and this is the main thing this round must settle: promote gives a task a new herdr pane, so any state anchoring a bound or a position against the OLD pane is invalid the instant that happens - carrying it either fires a bound on an innocent new pane or suppresses one that should fire, in both directions, silently. Enumerate every field with that property in both internal/state.Task (durable, on disk) and internal/watcher.TaskState (the live watcher's in-memory cache) - read the structs, do not assume a list - and classify each as: re-stamped on promote, reset to zero on promote, or genuinely pane-independent and therefore correctly carried. Pay particular attention to status_changed_for (added in a recent commit to fix this exact class - check whether it itself survives a promote uncorrected), and to report_offset (a promoted task keeps its id and therefore its status file, so carrying may be correct there - confirm and state why explicitly rather than assuming). Findings should name the field and the classification, not just flag that something is wrong.

Closes atqamz/secondhand#75

## What Changed

- Adds `hand watch --until-event` (with an optional `--timeout`): the watcher takes two silent baseline ticks, then writes the first tick's events to stdout and exits, so the exit itself is the delivery. Arming is bounded too - `connect` and a new `probeAllTasks` sweep race the context, exiting 4 when the window closes during arming or on a signal and 5 naming the one task whose pane could not be probed. `--timeout` without `--until-event` is rejected with exit 2.
- Adds a `parked` event for a worker that has gone silent with no herdr transition at all, with `config/parked-paused-bound` and `config/parked-other-bound` bounds (tasks whose last report was `done`/`failed` are exempt), a dashboard row, and a shared `configSeconds` helper replacing three copies of the read/Atoi/wrap block in `cmd/watch.go`.
- Makes the dwell and report bookkeeping durable in `state.Task` (`status_changed_at`, `status_changed_for`, `last_report_state`, `last_report_note`) so a watcher re-armed on every event resumes a real dwell instead of reseeding it to now, and treats a new herdr pane as invalidating everything anchored to the old one: `hand promote` restamps `status_changed_at` and clears the pane-scoped report/verified-done state, while `forgetPaneScopedCache` drops the cached status, dwell, stale/blocked/probe latches and last-report values on a pane-ID change (checked both at tick top and inside `syncTaskState` for a promote landing mid-tick), with the report file's mtime floored at the pane-start instant so a seconds-old ship does not inherit the scout's silence.

## Risk Assessment

Medium: Every defect from the previous rounds is closed on this head with matching regression tests, and the three intent criteria hold, but the accepted parked floor quietly redefines what parked measures (silence since the last pane transition, not since the last report) and SPECS now describes that anchor two different ways in the same file, which is worth correcting as a follow-up rather than blocking the merge.

## Testing

Ran the targeted watcher, cmd, and tagged e2e watch/promote tests under `nix develop` (all pass), then drove the real `hand` binary end-to-end through a 7-scene CLI transcript on a hermetic PATH and re-ran the identical driver against a binary built from the parent commit as a control. The control settles the round-1 finding: at the parent, scene F wrongly emits `idle-unreported ship-f` for a 6-second-old ship never observed working, and scene G stays silent while the ship's pane is blocked; at head both are correct (F silent, G announces `blocked ship-g`), and scenes A-E are byte-identical between the two runs modulo paths and timestamps, so the `ts.Status` reset changed exactly the two cases it targeted. The rest of the intent is visible in the head transcript: exit 4/5/2/0 with the exit-5 arm failure returning in 25ms of a 20s window, one config-seconds parser rejecting all three keys identically, `status_changed_for` resuming a 45m dwell in C1 and refusing a mismatched one in C2, promote clearing the pane-anchored durable fields while carrying `report_offset` with no replay, and the D6 negative control firing both bounds once the scout's stamps are pasted back. The three new unit tests are genuine regression coverage - they fail against the parent's product code and pass at head. This is a CLI-only change with no rendered surface, so the artifacts are command transcripts rather than screenshots. Worktree left clean; transient build and fake-home directories removed.

<details>
<summary>Evidence: Head (6f4cef5) end-to-end CLI transcript - all 7 scenes</summary>

```text

== build
   built /tmp/no-mistakes-evidence/01KYNWF6XFV6V9VGTQ05JFY2TM/work/hand from 6f4cef5 on HEAD

== ############ SCENE A - --until-event exit codes ############

== A1  quiet fleet: the window closes with nothing to report -> exit 4

$ hand watch --until-event --poll 200ms --timeout 2s
  stderr| no event within 2s
  exit  | 4
   ==> exit 4

== A2  one named worker whose pane cannot be probed -> exit 5, and it must not burn the window

$ hand watch --until-event --poll 200ms --timeout 20s
  stderr| could not arm: quiet-1: herdr pane get pane-q: exit status 1: herdr: pane pane-q not found
  exit  | 5
   ==> exit 5 after 25ms of a 20000ms window

== A3  --timeout without --until-event is a usage error -> exit 2

$ hand watch --timeout 5s
  stderr| --timeout requires --until-event: the streaming watcher has no completion to bound
  exit  | 2
   ==> exit 2

== A4  control: a real transition is delivered -> exit 0 with the event on stdout

$ hand watch --until-event --poll 200ms --timeout 30s
  stdout| idle-unreported quiet-1
  exit  | 0
   ==> exit 0

== ############ SCENE B - one config-seconds parser for all three keys ############

== B  config/stale-threshold

$ hand watch --until-event --timeout 2s
  stderr| invalid config/stale-threshold "not-a-number": strconv.Atoi: parsing "not-a-number": invalid syntax
  exit  | 1
   ==> exit 1

== B  config/parked-paused-bound

$ hand watch --until-event --timeout 2s
  stderr| invalid config/parked-paused-bound "not-a-number": strconv.Atoi: parsing "not-a-number": invalid syntax
  exit  | 1
   ==> exit 1

== B  config/parked-other-bound

$ hand watch --until-event --timeout 2s
  stderr| invalid config/parked-other-bound "not-a-number": strconv.Atoi: parsing "not-a-number": invalid syntax
  exit  | 1
   ==> exit 1

== ############ SCENE C - status_changed_for ties a resumed dwell to its status ############

== C1  durable stamp: working for 45m, pane still working -> the dwell resumes and stale fires

$ hand watch --poll 200ms   (streaming, stays up across the steps below)
  --- everything that watcher put on stdout:
  stdout| stale dwell-1
   state/dwell-1.json now carries the watcher's own stamp:
  json  |   "status_changed_at": "2026-07-29T03:58:26Z",
  json  |   "status_changed_for": "working",

== C2  same 45m stamp, recorded for a DIFFERENT status -> not trusted, dwell restarts, no stale

$ hand watch --poll 200ms   (streaming, stays up across the steps below)
  --- everything that watcher put on stdout:
  stdout| (no events)
   state/dwell-2.json re-stamped for the status actually observed:
  json  |   "status_changed_at": "2026-07-29T04:43:29Z",
  json  |   "status_changed_for": "working",

== ############ SCENE D - promote's durable state, and a watcher started after it ############

== D1  the scout's durable state before promote - every pane-anchored field populated
  json  | {
  json  |   "id": "scout-7",
  json  |   "project": "demo",
  json  |   "kind": "scout",
  json  |   "harness": "claude",
  json  |   "worktree": "/tmp/no-mistakes-evidence/01KYNWF6XFV6V9VGTQ05JFY2TM/work/wt-scout-7",
  json  |   "brief": "/tmp/no-mistakes-evidence/01KYNWF6XFV6V9VGTQ05JFY2TM/work/home/data/scout-7/brief.md",
  json  |   "herdr": {"session":"default","workspace_id":"ws-old","tab_id":"tab-old","pane_id":"pane-old"},
  json  |   "pr": "",
  json  |   "merged": false,
  json  |   "merged_at": "",
  json  |   "report_offset": 49,
  json  |   "pr_merged_observed": false,
  json  |   "done_verified": true,
  json  |   "created_at": "2026-07-29T03:53:32Z",
  json  |   "status_changed_at": "2026-07-29T03:58:32Z",
  json  |   "status_changed_for": "done",
  json  |   "last_report_state": "done",
  json  |   "last_report_note": "findings written"
  json  | }

== D2  hand promote

$ hand promote scout-7
  stdout| promoted scout-7: scout -> ship project=demo harness=claude
  exit  | 0
   ==> exit 0

== D3  the same task's durable state after promote
  json  | {
  json  |   "id": "scout-7",
  json  |   "project": "demo",
  json  |   "kind": "ship",
  json  |   "harness": "claude",
  json  |   "model": "",
  json  |   "effort": "",
  json  |   "worktree": "/tmp/no-mistakes-evidence/01KYNWF6XFV6V9VGTQ05JFY2TM/work/wt-ship-scout-7",
  json  |   "brief": "/tmp/no-mistakes-evidence/01KYNWF6XFV6V9VGTQ05JFY2TM/work/home/data/scout-7/brief.md",
  json  |   "herdr": {
  json  |     "session": "default",
  json  |     "workspace_id": "ws-new",
  json  |     "tab_id": "tab-new",
  json  |     "pane_id": "pane-new"
  json  |   },
  json  |   "pr": "",
  json  |   "merged": false,
  json  |   "merged_at": "",
  json  |   "report_offset": 49,
  json  |   "pr_merged_observed": false,
  json  |   "done_verified": false,
  json  |   "created_at": "2026-07-29T03:53:32Z",
  json  |   "status_changed_at": "2026-07-29T04:43:32Z",
  json  |   "status_changed_for": "",
  json  |   "last_report_state": "",
  json  |   "last_report_note": ""
  json  | }

== D4  a watcher started after the promote inherits no scout bound: stale (10m) and
   parked (60s) both start at the promotion, and the ship's own blocked still fires

$ hand watch --poll 200ms   (streaming, stays up across the steps below)
  --- everything that watcher put on stdout:
  stdout| blocked scout-7: agent needs help

== D5  report_offset was carried: the scout's consumed lines are not replayed, a new one flows

$ hand watch --poll 200ms   (streaming, stays up across the steps below)
  --- everything that watcher put on stdout:
  stdout| working scout-7: ship run started

== D6  negative control - give that same new pane the scout's stamps back
  json  |   "status_changed_at": "2026-07-29T03:58:41Z",
  json  |   "status_changed_for": "working",

$ hand watch --poll 200ms   (streaming, stays up across the steps below)
  --- everything that watcher put on stdout:
  stdout| stale scout-7
  stdout| parked scout-7: working: ship run started (silent 45m)

== ############ SCENE E - a promote landing UNDER a live watcher ############
   the in-memory TaskState survives the promote (created_at is unchanged), so only
   forgetPaneScopedCache stands between the scout's cached latches and the ship.

$ hand watch --poll 300ms   (streaming, stays up across the steps below)
   [t=2s] the scout has been done for 45m, so its own stale bound fires here
   [t=3s] hand promote scout-9 (watcher still up)
  promote| promoted scout-9: scout -> ship project=demo harness=claude
   [t=5s] the ship's pane goes blocked -> its own latch was re-armed, so this fires
   [t=7s] the ship reports done with no landing evidence yet -> reported-done, unverified
   [t=9s] the merge lands (merged=true, as hand merge would write it) -> the ship's
          own verified done fires, which a carried-over scout done_verified would have eaten
  --- everything that watcher put on stdout:
  stdout| stale scout-9
  stdout| blocked scout-9: agent needs help
  stdout| reported-done scout-9: shipped it
  stdout| done scout-9: shipped it
   no second 'stale scout-9' and no 'parked scout-9' after the promote line is the point:
   each of those bounds restarted with the new pane.

== ############ SCENE F - the cached herdr status is forgotten too ############
   TaskState.Status is the baseline ClassifyStatus compares the next probe against.
   It is cached against the OLD pane, so a pane change must reset it to herdr's
   'unknown' sentinel or the first probe of the NEW pane is diffed against the
   scout's status. This scene asserts the false-positive direction is gone.
   Setup: --poll 4s leaves the cache one interval stale, which is the window an
   operator hits by promoting off a live 'hand status' instead of a watch event.

$ hand watch --poll 4s   (streaming, stays up across the steps below)
   [t=6s] watcher has cached the scout pane as 'working'
   [t=6s] the scout finishes and is promoted inside one poll interval
  promote| promoted ship-f: scout -> ship project=demo harness=claude
   [t=6s] the brand-new ship pane answers 'done' on its first probe - a launched
          harness whose agent has not started working yet
  --- everything that watcher put on stdout:
  stdout| (no events)
   the ship is 6s old, was never observed working, and has no report of its own.
   PASS = no 'idle-unreported ship-f' line above. Any such line is the scout's
   status leaking across the pane boundary.

== ############ SCENE G - the same cached Status, the suppression direction ############
   Same root cause, other direction: if the new pane's first status happens to
   equal the scout's cached one, an un-reset baseline makes ClassifyStatus return
   early and the ship's blocked is never announced. Scene E is the control - there
   the cached status was 'done', so the ship's blocked fired regardless.

$ hand watch --poll 4s   (streaming, stays up across the steps below)
   [t=6s] watcher has cached the scout pane as 'blocked'; it then goes quiet and is
          promoted inside one poll interval
  promote| promoted ship-g: scout -> ship project=demo harness=claude
   [t=6s] the fresh ship pane is blocked - a worker asking its first question
  --- everything that watcher put on stdout:
  stdout| blocked ship-g: agent needs help
   PASS = a 'blocked ship-g' line above. Silence means the scout's cached status
   swallowed the ship's own question.

== done
```
</details>
<details>
<summary>Evidence: Parent (60b3e0b) control transcript - same driver, shows both bugs live</summary>

```text

== build
fatal: not a git repository (or any of the parent directories): .git
   built /tmp/no-mistakes-evidence/01KYNWF6XFV6V9VGTQ05JFY2TM/work-parent/hand from  on 

== ############ SCENE A - --until-event exit codes ############

== A1  quiet fleet: the window closes with nothing to report -> exit 4

$ hand watch --until-event --poll 200ms --timeout 2s
  stderr| no event within 2s
  exit  | 4
   ==> exit 4

== A2  one named worker whose pane cannot be probed -> exit 5, and it must not burn the window

$ hand watch --until-event --poll 200ms --timeout 20s
  stderr| could not arm: quiet-1: herdr pane get pane-q: exit status 1: herdr: pane pane-q not found
  exit  | 5
   ==> exit 5 after 26ms of a 20000ms window

== A3  --timeout without --until-event is a usage error -> exit 2

$ hand watch --timeout 5s
  stderr| --timeout requires --until-event: the streaming watcher has no completion to bound
  exit  | 2
   ==> exit 2

== A4  control: a real transition is delivered -> exit 0 with the event on stdout

$ hand watch --until-event --poll 200ms --timeout 30s
  stdout| idle-unreported quiet-1
  exit  | 0
   ==> exit 0

== ############ SCENE B - one config-seconds parser for all three keys ############

== B  config/stale-threshold

$ hand watch --until-event --timeout 2s
  stderr| invalid config/stale-threshold "not-a-number": strconv.Atoi: parsing "not-a-number": invalid syntax
  exit  | 1
   ==> exit 1

== B  config/parked-paused-bound

$ hand watch --until-event --timeout 2s
  stderr| invalid config/parked-paused-bound "not-a-number": strconv.Atoi: parsing "not-a-number": invalid syntax
  exit  | 1
   ==> exit 1

== B  config/parked-other-bound

$ hand watch --until-event --timeout 2s
  stderr| invalid config/parked-other-bound "not-a-number": strconv.Atoi: parsing "not-a-number": invalid syntax
  exit  | 1
   ==> exit 1

== ############ SCENE C - status_changed_for ties a resumed dwell to its status ############

== C1  durable stamp: working for 45m, pane still working -> the dwell resumes and stale fires

$ hand watch --poll 200ms   (streaming, stays up across the steps below)
  --- everything that watcher put on stdout:
  stdout| stale dwell-1
   state/dwell-1.json now carries the watcher's own stamp:
  json  |   "status_changed_at": "2026-07-29T03:59:56Z",
  json  |   "status_changed_for": "working",

== C2  same 45m stamp, recorded for a DIFFERENT status -> not trusted, dwell restarts, no stale

$ hand watch --poll 200ms   (streaming, stays up across the steps below)
  --- everything that watcher put on stdout:
  stdout| (no events)
   state/dwell-2.json re-stamped for the status actually observed:
  json  |   "status_changed_at": "2026-07-29T04:44:59Z",
  json  |   "status_changed_for": "working",

== ############ SCENE D - promote's durable state, and a watcher started after it ############

== D1  the scout's durable state before promote - every pane-anchored field populated
  json  | {
  json  |   "id": "scout-7",
  json  |   "project": "demo",
  json  |   "kind": "scout",
  json  |   "harness": "claude",
  json  |   "worktree": "/tmp/no-mistakes-evidence/01KYNWF6XFV6V9VGTQ05JFY2TM/work-parent/wt-scout-7",
  json  |   "brief": "/tmp/no-mistakes-evidence/01KYNWF6XFV6V9VGTQ05JFY2TM/work-parent/home/data/scout-7/brief.md",
  json  |   "herdr": {"session":"default","workspace_id":"ws-old","tab_id":"tab-old","pane_id":"pane-old"},
  json  |   "pr": "",
  json  |   "merged": false,
  json  |   "merged_at": "",
  json  |   "report_offset": 49,
  json  |   "pr_merged_observed": false,
  json  |   "done_verified": true,
  json  |   "created_at": "2026-07-29T03:55:02Z",
  json  |   "status_changed_at": "2026-07-29T04:00:02Z",
  json  |   "status_changed_for": "done",
  json  |   "last_report_state": "done",
  json  |   "last_report_note": "findings written"
  json  | }

== D2  hand promote

$ hand promote scout-7
  stdout| promoted scout-7: scout -> ship project=demo harness=claude
  exit  | 0
   ==> exit 0

== D3  the same task's durable state after promote
  json  | {
  json  |   "id": "scout-7",
  json  |   "project": "demo",
  json  |   "kind": "ship",
  json  |   "harness": "claude",
  json  |   "model": "",
  json  |   "effort": "",
  json  |   "worktree": "/tmp/no-mistakes-evidence/01KYNWF6XFV6V9VGTQ05JFY2TM/work-parent/wt-ship-scout-7",
  json  |   "brief": "/tmp/no-mistakes-evidence/01KYNWF6XFV6V9VGTQ05JFY2TM/work-parent/home/data/scout-7/brief.md",
  json  |   "herdr": {
  json  |     "session": "default",
  json  |     "workspace_id": "ws-new",
  json  |     "tab_id": "tab-new",
  json  |     "pane_id": "pane-new"
  json  |   },
  json  |   "pr": "",
  json  |   "merged": false,
  json  |   "merged_at": "",
  json  |   "report_offset": 49,
  json  |   "pr_merged_observed": false,
  json  |   "done_verified": false,
  json  |   "created_at": "2026-07-29T03:55:02Z",
  json  |   "status_changed_at": "2026-07-29T04:45:02Z",
  json  |   "status_changed_for": "",
  json  |   "last_report_state": "",
  json  |   "last_report_note": ""
  json  | }

== D4  a watcher started after the promote inherits no scout bound: stale (10m) and
   parked (60s) both start at the promotion, and the ship's own blocked still fires

$ hand watch --poll 200ms   (streaming, stays up across the steps below)
  --- everything that watcher put on stdout:
  stdout| blocked scout-7: agent needs help

== D5  report_offset was carried: the scout's consumed lines are not replayed, a new one flows

$ hand watch --poll 200ms   (streaming, stays up across the steps below)
  --- everything that watcher put on stdout:
  stdout| working scout-7: ship run started

== D6  negative control - give that same new pane the scout's stamps back
  json  |   "status_changed_at": "2026-07-29T04:00:10Z",
  json  |   "status_changed_for": "working",

$ hand watch --poll 200ms   (streaming, stays up across the steps below)
  --- everything that watcher put on stdout:
  stdout| stale scout-7
  stdout| parked scout-7: working: ship run started (silent 45m)

== ############ SCENE E - a promote landing UNDER a live watcher ############
   the in-memory TaskState survives the promote (created_at is unchanged), so only
   forgetPaneScopedCache stands between the scout's cached latches and the ship.

$ hand watch --poll 300ms   (streaming, stays up across the steps below)
   [t=2s] the scout has been done for 45m, so its own stale bound fires here
   [t=3s] hand promote scout-9 (watcher still up)
  promote| promoted scout-9: scout -> ship project=demo harness=claude
   [t=5s] the ship's pane goes blocked -> its own latch was re-armed, so this fires
   [t=7s] the ship reports done with no landing evidence yet -> reported-done, unverified
   [t=9s] the merge lands (merged=true, as hand merge would write it) -> the ship's
          own verified done fires, which a carried-over scout done_verified would have eaten
  --- everything that watcher put on stdout:
  stdout| stale scout-9
  stdout| blocked scout-9: agent needs help
  stdout| reported-done scout-9: shipped it
  stdout| done scout-9: shipped it
   no second 'stale scout-9' and no 'parked scout-9' after the promote line is the point:
   each of those bounds restarted with the new pane.

== ############ SCENE F - the cached herdr status is forgotten too ############
   TaskState.Status is the baseline ClassifyStatus compares the next probe against.
   It is cached against the OLD pane, so a pane change must reset it to herdr's
   'unknown' sentinel or the first probe of the NEW pane is diffed against the
   scout's status. This scene asserts the false-positive direction is gone.
   Setup: --poll 4s leaves the cache one interval stale, which is the window an
   operator hits by promoting off a live 'hand status' instead of a watch event.

$ hand watch --poll 4s   (streaming, stays up across the steps below)
   [t=6s] watcher has cached the scout pane as 'working'
   [t=6s] the scout finishes and is promoted inside one poll interval
  promote| promoted ship-f: scout -> ship project=demo harness=claude
   [t=6s] the brand-new ship pane answers 'done' on its first probe - a launched
          harness whose agent has not started working yet
  --- everything that watcher put on stdout:
  stdout| idle-unreported ship-f
   the ship is 6s old, was never observed working, and has no report of its own.
   PASS = no 'idle-unreported ship-f' line above. Any such line is the scout's
   status leaking across the pane boundary.

== ############ SCENE G - the same cached Status, the suppression direction ############
   Same root cause, other direction: if the new pane's first status happens to
   equal the scout's cached one, an un-reset baseline makes ClassifyStatus return
   early and the ship's blocked is never announced. Scene E is the control - there
   the cached status was 'done', so the ship's blocked fired regardless.

$ hand watch --poll 4s   (streaming, stays up across the steps below)
   [t=6s] watcher has cached the scout pane as 'blocked'; it then goes quiet and is
          promoted inside one poll interval
  promote| promoted ship-g: scout -> ship project=demo harness=claude
   [t=6s] the fresh ship pane is blocked - a worker asking its first question
  --- everything that watcher put on stdout:
  stdout| (no events)
   PASS = a 'blocked ship-g' line above. Silence means the scout's cached status
   swallowed the ship's own question.

== done
```
</details>
<details>
<summary>Evidence: Before/after signal summary</summary>

<code>Same driver (drive.sh), two binaries, only the promote-forget differs.&#10;&#10;parent 60b3e0b head 6f4cef5&#10;scene F (false positive) idle-unreported ship-f (no events) &lt;- fixed&#10;scene G (suppression) (no events) blocked ship-g &lt;- fixed&#10;scenes A-E identical identical&#10;&#10;Unit-level control: HEAD&#39;s internal/watcher/watcher_test.go compiled against parent&#10;product code fails all three pane-status assertions, and passes at HEAD.&#10;TestForgetPaneScopedCacheClearsEveryPaneAnchoredLatch Status = &#34;working&#34;&#10;TestForgetPaneScopedCacheStopsIdleUnreportedForAStatusTheShipNeverHeld fired&#10;TestForgetPaneScopedCacheLetsTheShipsOwnBlockedFireAfterABlockedScout nil</code>

```text
Same driver (drive.sh), two binaries, only the promote-forget differs.

                         parent 60b3e0b            head 6f4cef5
scene F (false positive) idle-unreported ship-f     (no events)      <- fixed
scene G (suppression)    (no events)                blocked ship-g   <- fixed
scenes A-E               identical                  identical

scene A  exit 4 quiet window / exit 5 named arm-probe failure in 25ms of a 20s window
         / exit 2 --timeout without --until-event / exit 0 with the event on stdout
scene B  one config-seconds parser rejects all three keys the same way
scene C  status_changed_for: C1 resumes a 45m dwell and fires stale, C2 restarts it
scene D  promote clears status_changed_for/last_report_*/done_verified, carries
         report_offset (D5: no replay, a new line flows), D6 negative control fires
         stale + parked once the scout's stamps are pasted back
scene E  promote under a live watcher: stale before, blocked/reported-done/done after,
         no inherited second stale and no inherited parked

Unit-level control: HEAD's internal/watcher/watcher_test.go compiled against parent
product code fails all three pane-status assertions, and passes at HEAD.
  TestForgetPaneScopedCacheClearsEveryPaneAnchoredLatch          Status = "working"
  TestForgetPaneScopedCacheStopsIdleUnreportedForAStatusTheShipNeverHeld   fired
  TestForgetPaneScopedCacheLetsTheShipsOwnBlockedFireAfterABlockedScout    nil
```
</details>
<details>
<summary>Evidence: Pane-anchored field classification for state.Task and watcher.TaskState (Status row updated to reset-on-promote)</summary>

```text
# Every field that could anchor a bound or a position against a pane `hand promote` replaced

Read from the structs at target commit `60b3e0b`, then each classification exercised
against the built `hand` binary - see `transcript.txt`, scenes referenced per row.
`drive.sh` reproduces the whole transcript.

`hand promote` keeps the task's `id` and `created_at`, so neither the durable state file
nor the watcher's in-memory `TaskState` is discarded: the identity check in
`internal/watcher/watcher.go:193` deliberately does not fire. Everything below is what
that leaves to be corrected explicitly.

## internal/state.Task (durable, state/<id>.json)

| field | classification | where |
|---|---|---|
| `ID` | carried - identity, and what makes `state/<id>.status` continuous | - |
| `Project` | carried - pane-independent | - |
| `Kind` | rewritten scout -> ship | promote.go:164 |
| `Harness` / `Model` / `Effort` | rewritten from flags/brief front matter | promote.go:165-167 |
| `Worktree` | rewritten to the new lease; old one returned | promote.go:168,194 |
| `Brief` | carried - same brief drives the ship | - |
| `Herdr.{Session,WorkspaceID,TabID,PaneID}` | rewritten - this *is* the new pane | promote.go:169-174 |
| `PR` | carried - pane-independent (a PR is repo state, not pane state) | - |
| `MergeExecuted` / `MergeExecutedAt` | carried - pane-independent; a scout has none | - |
| `MergeAnnounced` (`pr_merged_observed`) | carried - PR-scoped, not pane-scoped | - |
| `ReportOffset` | **carried, and correct**: the report channel is `state/<id>.status`, keyed by the id promote keeps, and promote never truncates or rotates it. Resetting to 0 would replay every scout line the operator already saw; carrying it keeps one continuous stream | promote.go:176-177, scene D5 |
| `DoneVerified` | reset to `false` | promote.go:179, scene E |
| `CreatedAt` | carried by design - the watcher's respawn check keys on it | - |
| `StatusChangedAt` | **re-stamped to now** | promote.go:180, scenes D3/D4 vs D6 |
| `StatusChangedFor` | **reset to `""`** - it does *not* survive uncorrected | promote.go:181, scene D4 vs D6 |
| `LastReportState` / `LastReportNote` | reset to `""` | promote.go:182-183 |

The `StatusChangedFor = ""` reset is load-bearing twice over, and the two uses want
opposite things from the same restamped `StatusChangedAt`:

- `statusChangeSeed` (watcher.go:334) sees a non-empty stamp whose `for` matches no
  status, so it returns `now` - the ship's dwell starts at its own first observation,
  never at the scout's (scene D4: no `stale`; scene D6, the same pane handed the scout's
  stamps back, fires `stale` immediately).
- `paneStartTime` (watcher.go:375) wants exactly the restamped value, as the floor
  `reportEvidenceTime` puts under the report mtime, so a 45-minute-old scout channel
  cannot make a seconds-old ship `parked` (scene D4: no `parked`; scene D6: `parked ...
  (silent 45m)`).

## internal/watcher.TaskState (live, in-memory)

`forgetPaneScopedCache` (watcher.go:274) triggers on `PersistedPaneID != t.Herdr.PaneID`,
and is called both at the top of `tick` and again inside `syncTaskState` for a promote
that lands mid-tick.

| field | classification | where |
|---|---|---|
| `CreatedAt` | carried by design - promote keeps `created_at`, so tracking must continue | watcher.go:193 |
| `Status` | **reset to `herdr.StatusUnknown`** - see note below | watcher.go:295, scenes F/G |
| `Probed` | **reset to `true`** - matches `NewTaskState`'s unconditional seed for a first sighting, so an unresolved probe failure inherited from the old pane cannot swallow the ship's own first probe failure as no edge, and does not leave `ClassifyStale`'s early return gated off | watcher.go:303 |
| `ChangedAt` | re-seeded via `statusChangeSeed` | watcher.go:287-288, scene E |
| `Blocked` | reset to `false` | watcher.go:292, scene E |
| `Stale` | reset to `false` | watcher.go:291, scene E |
| `PRMerged` | carried - PR-scoped | - |
| `ReportOffset` / `PersistedOffset` | carried - id-scoped, same reason as the durable field | scene D5 |
| `PersistedPRMerged` | carried - mirror of a PR-scoped fact | - |
| `DoneVerified` / `PersistedDoneVerified` | reset when the disk value went false | watcher.go:279-282, scene E |
| `PersistedPaneID` | updated to the new pane - the trigger itself | watcher.go:286 |
| `PersistedChangedAt` | reset to the fresh seed | watcher.go:289 |
| `PersistedChangedFor` | re-mirrored from disk (`""`), so the next write restamps it | watcher.go:290 |
| `LastReportState` / `LastReportNote` | reset from disk (`""` after promote) | watcher.go:293-294 |
| `ParkedFiredFor` | carried, and safe: it holds a *floored* report mtime, and `reportEvidenceTime`'s floor is the new pane's start, which is strictly later than any value the scout's pane could have latched - so the ship's first parked check can never match it | watcher.go:354-370, scenes D4/E |

## Note: the forget trigger compares pane identity, not a timestamp proxy

An earlier round of this same fix compared `t.StatusChangedAt` against `ts.PersistedChangedAt`
as the promote signal, on the theory that a promote's restamp would show up as a divergence.
That string could not carry the weight: `statusChangeSeed` returns `now` for reasons that have
nothing to do with promote - any resume where the arm-time observed status disagrees with the
durable `status_changed_for` - so the comparison false-positived on ordinary `--until-event`
resumption, and separately false-negatived whenever a real promote's restamp landed in the same
RFC3339-seconds-granularity wall-clock second as what was already persisted. `PersistedPaneID`
does not have either failure mode: the herdr pane ID is the one thing `hand promote` actually
replaces to create the ship, so comparing it against the cached value detects a promote exactly
when one happened and never when one did not, independent of timing.

## Note: `TaskState.Status` is pane-anchored, and is now reset (verified)

`Status` is the baseline `ClassifyStatus` diffs the next probe against
(`internal/watcher/events.go:137`). Carrying it across a promote compares the first probe
of the **new** pane against a status that belonged to a different agent, and both failure
directions were reproduced against a binary built from the parent commit `60b3e0b`
(`transcript-parent-control.txt`):

- **fires on an innocent new pane** (scene F): cached scout status `working`, ship's first
  probe reads not-busy, `LastReportState` just cleared by the forget -> `idle-unreported
  ship-f` for a ship 6 seconds old that was never observed working.
- **suppresses one that should fire** (scene G): cached scout status `blocked`, ship's
  first probe also reads `blocked` -> `status == ts.Status` returns early, so `Blocked`
  is never re-latched and the ship's blocked question is announced never. Scene E is the
  control: same promote, cached status `done`, and there the ship's `blocked` does fire.

Reachability: the cache only has to disagree with the pane by one poll interval, which is
the normal case for an operator who promotes off a live `hand status` rather than off a
watch event, and is unconditional for a promote landing inside a tick (`syncTaskState`
calls the forget with `ts.Status` already set from the old pane's probe *this* tick).

`6f4cef5` sets `ts.Status = herdr.StatusUnknown` in the pane-change branch, after the
`statusChangeSeed` call that still wants to know what status the cached dwell described.
`StatusUnknown` matches neither the `working` nor the `blocked` branch, so the ship's
first probe reads as a baseline rather than as a transition out of the scout's status.
On the same driver at `6f4cef5` (`transcript.txt`) scene F is silent and scene G announces
`blocked ship-g`; scenes A-E are byte-identical to the parent run modulo paths and
timestamps, so the reset changed exactly the two cases it was aimed at.
```
</details>
- Evidence: Hermetic end-to-end driver used for both runs (local file: <code>/tmp/no-mistakes-evidence/01KYNWF6XFV6V9VGTQ05JFY2TM/drive.sh</code>)
<details>
<summary>Evidence: Scene A/G decisive lines from the two transcripts</summary>

```text
parent 60b3e0b, scene F:
--- everything that watcher put on stdout:
stdout| idle-unreported ship-f
parent 60b3e0b, scene G:
--- everything that watcher put on stdout:
stdout| (no events)

head 6f4cef5, scene F:
--- everything that watcher put on stdout:
stdout| (no events)
head 6f4cef5, scene G:
--- everything that watcher put on stdout:
stdout| blocked ship-g: agent needs help
```
</details>
- Outcome: 1 issue found -> auto-fixed across 2 runs (27m59s)

## Pipeline

Updates from [git push no-mistakes](https://github.com/kunchenguid/no-mistakes)

<details>
<summary>**intent** - passed</summary>

No issues found.

</details>

<details>
<summary>**Rebase** - passed</summary>

No issues found.

</details>

<details>
<summary>**Review** - 3 infos</summary>

- warning: `internal/watcher/watcher.go:224` - internal/watcher/watcher.go:224 - the promote divergence branch reseeds ChangedAt/PersistedChangedAt but leaves ts.Stale latched, so the reseeded dwell is inert in exactly the scenario the branch&#39;s own comment names. Classification: ts.Stale is pane-anchored and is neither reset nor re-stamped on promote. Sequence: (1) scout dwells past StaleThreshold, watcher fires `stale &lt;id&gt;` and sets ts.Stale=true; (2) operator runs `hand promote`, disk status_changed_at is restamped; (3) next tick detects divergence and resets ChangedAt to promotion time, but the ship&#39;s first probe reads the same herdr status the scout last held (the case the comment cites verbatim), so ClassifyStatus returns at events.go:132 before reaching `ts.Stale = false`; (4) the ship&#39;s worker hangs with no further herdr transition - ClassifyStale bails on `ts.Stale` at events.go:160 and the stale event never fires for the new pane. Fix: clear ts.Stale (and ts.Blocked, same class, same latch-suppression shape) inside the divergence branch alongside the reseed.
- warning: `internal/watcher/watcher.go:554` - internal/watcher/watcher.go:554 - syncTaskState re-reads the task under the lock but then unconditionally writes the cached ChangedAt (and ORs the cached DoneVerified) back, so a promote that lands mid-tick is silently undone and the tick-top forget rules can never see it. Sequence: tick calls state.List at watcher.go:172, snapshotting the pre-promote task; promote acquires the task lock and completes during that same tick (plausible - a tick can span tens of seconds when `gh pr view` runs for earlier tasks, capped at 30s each); syncTaskState then wins the TryLock, re-reads the freshly promoted task, and overwrites t.StatusChangedAt with the scout-era ts.ChangedAt (and resurrects done_verified via the OR at line 553). On the following tick the disk value now equals ts.PersistedChangedAt, so the divergence check at line 224 sees nothing to forget and the ship carries the scout&#39;s dwell for the rest of the run. The earliest shared boundary is syncTaskState&#39;s own lock-protected re-read: compare the re-read t.StatusChangedAt / t.DoneVerified against ts.Persisted* there and drop the cached values instead of writing them back, rather than adding a third symptom check at the top of tick.
- warning: `cmd/promote.go:183` - cmd/promote.go:183 - promote re-stamps status_changed_at and clears done_verified but leaves the scout&#39;s report evidence as the ship&#39;s &#34;last report&#34;, disabling two watchdogs and re-arming a third for the promoted task. LastReportState/LastReportNote are report-file anchored, so they are continuous by construction (and resumeTaskState re-derives the same scout line from disk at watcher.go:298) - but the line was written by the old worker in the old pane. Consequences until the ship writes its first report line: (a) ClassifyParked exempts the task entirely, since parkedBound treats the scout&#39;s `done` as terminal (events.go:182); (b) ClassifyStatus absorbs the ship pane&#39;s unexplained stop instead of firing idle-unreported, because LastReportState is neither empty nor `working` (events.go:144); (c) ClassifyDeferredDone re-fires - ts.DoneVerified was reset but ts.LastReportState is still `done`, so if the scout had a recorded PR the watcher saw merged (cmd/pr.go has no scout gate, and autoRecordPR at watcher.go:325 has no kind check either), doneVerified&#39;s ship branch `t.PR != &#34;&#34; &amp;&amp; ts.PRMerged` is true and the watcher announces a verified `done &lt;id&gt;` with the scout&#39;s note on the ship&#39;s first tick. Single shared fix: have promote append one line to state/&lt;id&gt;.status (e.g. `working: promoted to ship`), which advances the mtime the parked clock is anchored to and reclassifies LastReportState through the normal tail path for both the live and the resumed watcher.
- warning: `internal/watcher/watcher.go:165` - internal/watcher/watcher.go:165 - probeAllTasks returns ErrArmFailed (exit 5) when ctx expires or is signaled during the arm probe, contradicting the stated criterion &#34;the exit-code fix distinguishing arm-timeout/signal (exit 4) from a named per-task probe failure (exit 5)&#34; and this change&#39;s own docs: SPECS.md step 5 says &#34;On SIGINT/SIGTERM, the same as a timeout: nothing was delivered, so exit 4, never 0&#34;, and AGENTS.md/README define exit 5 as &#34;a worker named on stderr&#34;. With a wedged herdr whose PaneGet hangs, `hand watch --until-event --timeout 30m` exits 5 with &#34;could not arm: timed out probing tasks&#34; and no worker named; Ctrl-C during the same probe exits 5 with the same message claiming a timeout that did not occur. Only the per-task branch at line 157 should carry ErrArmFailed; the ctx branch should wrap ErrNoEvent (exit 4) and distinguish DeadlineExceeded from Canceled in its text. Note the sibling case in connect at line 135 is documented as deliberately exit 1, so only probeAllTasks needs changing.
- info: `cmd/watch.go:62` - cmd/watch.go:62 - the config-seconds dedup named in the intent is not present on this head: stale-threshold (line 54), parked-paused-bound (line 62) and parked-other-bound (line 71) are three verbatim copies of the same read/Atoi/multiply/wrap-error block, and the two new ones added it twice more. A single helper - `func configSeconds(home, name string, fallback time.Duration) (time.Duration, error)` next to configDefault in cmd/spawn.go:237 - collapses all three with no behavior change.
- info: `cmd/promote.go:175` - cmd/promote.go:175 - confirming the report_offset carry rather than assuming it: state.ReportPath (internal/state/report.go:36) derives the path from the task id alone, promote keeps the id and never opens or truncates state/&lt;id&gt;.status, and TailReport treats the file as append-only. So the persisted offset points at the byte immediately after the scout&#39;s last consumed line, which is exactly where the ship&#39;s first appended line lands - carrying it is correct, and zeroing it would replay every scout line the durable offset exists to prevent replaying. ParkedFiredFor is correct for the same reason: it is keyed on that file&#39;s mtime, not on the pane. No action needed.

Fix: forget every pane-scoped cached fact across promote
5 issues (1 error, 3 warnings, 1 info) still open:
- error: `internal/watcher/watcher.go:532` - forgetPaneScopedCache is triggered by `t.StatusChangedAt != ts.PersistedChangedAt` (watcher.go:274), which diverges with no promote at all whenever statusChangeSeed returned `now` at resume - that is, whenever the arm-time observed status differs from the durable status_changed_for. resumeTaskState (watcher.go:243-246) seeds PersistedChangedAt=now while the disk stamp stays old, so the divergence is true on the first classifying tick. The sync-time call at watcher.go:532 runs AFTER tailReport and reverts ts.LastReportState/ts.LastReportNote to the disk values (watcher.go:283-284), then persists them alongside the advanced report_offset (watcher.go:534-540). Since ruling item 3 made resumeTaskState stop re-reading state/&lt;id&gt;.status, the consumed line&#39;s state is unrecoverable. Fix: trigger the forget on pane identity (cache t.Herdr.PaneID in TaskState and compare it) rather than on timestamp divergence - that is the invariant promote actually changes, and it has no false positive from the watcher&#39;s own resume seeding.
- warning: `internal/watcher/watcher.go:123` - Intent criterion: &#34;the exit-code fix distinguishing arm-timeout/signal (exit 4) from a named per-task probe failure (exit 5) in connect/probeAllTasks&#34;. It holds in probeAllTasks (watcher.go:152-157) but not in connect: its ctx branch returns `fmt.Errorf(&#34;herdr unreachable: %w&#34;, ctx.Err())`, which wraps neither ErrNoEvent nor ErrArmFailed, so cmd/watch.go:85-92 falls through to exit 1. SPECS.md now states the opposite for the whole arm phase - step 1 lists connect as arming and says both halves are bounded by --timeout, step 4 says &#34;On --timeout, whether it elapses during arming or during the poll ... exit 4&#34;, and step 5 says a signal is always 4, never a watcher failure. Whether connect&#39;s ctx race should become ErrNoEvent (exit 4) or stay a genuine watcher failure (exit 1, with SPECS narrowed to the probe sweep) is a design call, so raising rather than fixing.
- warning: `internal/watcher/watcher.go:339` - The parked bound is anchored to the report file&#39;s mtime (reportEvidenceTime, watcher.go:339-349), which promote deliberately leaves untouched, while promote now clears last_report_state (cmd/promote.go:182-183). parkedBound (events.go:175-188) therefore flips the task from exempt (scout&#39;s &#34;done&#34;) to the default Other bound measured against the scout&#39;s last append. This is the &#34;fires a bound on an innocent new pane&#34; direction, and it is a regression: on the previous heads the watcher re-derived &#34;done&#34; from the file and stayed exempt. Needs a durable pane-start anchor (spawn and promote both stamping it, reportEvidenceTime flooring mtime at it); status_changed_at cannot serve, since it moves on every transition and would suppress parked for a pane that flips status without ever reporting.
- warning: `internal/watcher/watcher.go:274` - The promote detector compares RFC3339 strings, which are second-granular. A promote whose restamp lands in the same wall-clock second as the value this watcher last persisted is invisible, so nothing is forgotten. This is the false-negative direction of the same class: the stale dwell, the Stale and Blocked latches, DoneVerified and LastReportState/Note all carry onto the ship&#39;s new pane, and the disk value then matches what the watcher persisted, so no later tick finds anything to forget either. Same fix as the report-line clobber: compare t.Herdr.PaneID against a cached copy in TaskState, which is exact and has no granularity to lose.
- info: `SPECS.md:763` - The new &#34;Pane-anchored facts across hand promote&#34; section claims a field-by-field classification, and the restart table above it states &#34;Anything added to TaskState belongs in this table before it ships&#34;, but neither table lists ParkedFiredFor or the report-file mtime the parked bound is measured against. ParkedFiredFor belongs in the carried table (it is keyed to the report mtime, not the pane, and promote leaves the report stream continuous, so suppressing a duplicate for an unchanged mtime stays correct). The mtime anchor itself needs a row once the false-parked issue is settled.

Fix: fix pane-identity cache trigger, parked floor, connect exit code
3 infos still open:
- info: `internal/watcher/watcher.go:361` - Recording the actual semantics of the accepted fix, since the ruling&#39;s premise (&#34;StatusChangedAt ... already carr[ies] this instant for promote&#34;) holds only at the moment of a promote. paneStartTime returns t.StatusChangedAt, which syncTaskState rewrites on every observed herdr transition (watcher.go:537), not just at pane creation. So reportEvidenceTime&#39;s floor advances whenever the pane changes status, and parked now measures &#34;silence since the last pane activity&#34; rather than &#34;silence on the report channel&#34;. Consequences: for a genuinely stuck worker (no transitions) the bound is only delayed, measured from when it entered its current status; for a worker that keeps flipping status without ever appending to state/&lt;id&gt;.status, parked never fires at all. That second case is not a silent hole - each flip that stops the pane raises idle-unreported through ClassifyStatus - and the round-2 objection to using status_changed_at was explicitly overridden, so no action. Flagging only so the record matches the code: SPECS.md:770 calls the floor &#34;the pane-start instant&#34;, which is not what status_changed_at is outside the promote case.
- info: `SPECS.md:742` - Two gaps in the tables this change owns. (1) The restart table&#39;s parked row still reads &#34;What the bound is measured against - the report file&#39;s own mtime, or created_at for a task that has never reported&#34;, which this head no longer implements: reportEvidenceTime returns max(mtime, status_changed_at|created_at), and the new promote-section row at SPECS.md:770 says so, leaving the two rows contradicting each other in the same file. (2) SPECS.md:744 states &#34;Anything added to TaskState belongs in this table before it ships&#34;, and PersistedPaneID (internal/watcher/events.go:82) is a new TaskState field with no row - it is re-derived safely on resume from the durable t.Herdr.PaneID, which is exactly the kind of entry the table exists to record.
- info: `internal/watcher/watcher.go:223` - reportEvidenceTime now parses timestamps before it stats anything (watcher.go:350-353), so its error can be `parse status_changed_at &#34;...&#34;` while tick still prints it as &#34;watch: stat report %s failed: %v&#34;. A task whose stamps will not parse also has ClassifyParked skipped for it on every tick while that misattributed line repeats at the poll interval. Not reachable from state hand writes itself (cmd/spawn.go:169 always sets created_at, and status_changed_at is always written as RFC3339), so this is a diagnostics fix only: reword the message to name the evidence-time lookup rather than the stat.

</details>

<details>
<summary>**Test** - 1 issue found -> auto-fixed </summary>

- error: `internal/watcher/watcher.go:274` - internal/watcher.TaskState.Status is pane-anchored but is not reset by forgetPaneScopedCache (internal/watcher/watcher.go:274-295), so the first probe of the pane hand promote just created is diffed against the status the OLD pane held. Reproduced end-to-end against the built binary in both directions (transcript.txt scenes F and G). Fires wrongly: cached scout status `working`, ship&#39;s first probe reads not-busy, LastReportState just cleared by the forget -&gt; `idle-unreported ship-f` for a ship 6 seconds old that was never observed working. Suppresses wrongly: cached scout status `blocked`, ship&#39;s first probe also `blocked` -&gt; `status == ts.Status` returns early at internal/watcher/events.go:137, so ts.Blocked is never re-latched and the ship&#39;s blocked question is announced never (scene E is the control - same promote with cached status `done`, and there blocked does fire). The cache only has to disagree with the pane by one poll interval, which is the normal case for an operator promoting off a live `hand status` rather than off a watch event, and is unconditional for a promote landing mid-tick, where syncTaskState calls the forget with ts.Status already set from the old pane&#39;s probe in that same tick. Every other pane-anchored field in both structs is correctly re-stamped, reset, or provably safe to carry - see pane-scoped-field-classification.md for the full per-field table. The fix is product code (reset Status, or treat the first post-promote probe as a re-seed rather than a transition), so it is the author&#39;s call rather than a test-phase edit.
- `CGO_ENABLED=0 go build ./...`
- <code>`go test -race -count=1 ./internal/watcher/ ./cmd/` (234 tests)</code>
- `go test -tags e2e -race -count=1 ./tests/e2e/ -run 'TestWatchUntilEvent'`
- `go test -tags e2e -race -count=1 ./tests/e2e/ -run 'TestWatch|TestPromoteScoutToShip|TestReport'`
- <code>Manual CLI transcript via `bash /tmp/no-mistakes-evidence/01KYNWF6XFV6V9VGTQ05JFY2TM/drive.sh` against the built binary on a hermetic PATH</code>
- <code>scene A: `hand watch --until-event --timeout 2s` (exit 4), unprobeable pane with `--timeout 20s` (exit 5 in 27ms), `hand watch --timeout 5s` (exit 2), delivered transition (exit 0)</code>
- <code>signal path: SIGTERM to `hand watch --until-event` while herdr&#39;s reachability probe hangs (exit 4)</code>
- <code>scene B: `config/stale-threshold`, `config/parked-paused-bound`, `config/parked-other-bound` each set to a non-number and rejected by the shared parser</code>
- <code>scene C: 45m `status_changed_at` with matching vs mismatched `status_changed_for` under `hand watch --poll 200ms`</code>
- <code>scene D: `hand promote scout-7` with state/&lt;id&gt;.json dumped before and after, then a fresh watcher checking stale/parked/report-replay, with a stamps-carried-back negative control</code>
- <code>scene E: `hand promote` landing under a live `hand watch --poll 300ms`, then the ship&#39;s own blocked, reported-done and deferred verified done</code>
- `scenes F and G: promote inside one poll interval, new pane's first probe diffed against the scout's cached status (false idle-unreported; suppressed blocked)`

Fix: reset cached herdr status when promote changes pane
Re-checked - no issues remain.
- `nix develop -c go test -race -count=1 -run "ForgetPaneScoped|SyncTaskState|StatusChanged|ClassifyStatus|Promote" ./internal/watcher/... ./cmd/...`
- <code>`nix develop -c go test -race -count=1 ./internal/watcher/...` (full watcher package)</code>
- `nix develop -c go test -race -count=1 -run "Watch|UntilEvent|Promote" ./cmd/...`
- `nix develop -c go test -race -count=1 -tags e2e -run "Watch|Promote" ./tests/e2e/...`
- <code>End-to-end driver against the real `hand` binary built from 6f4cef5 on a hermetic PATH (fake herdr/treehouse, throwaway home): `REPO=&lt;worktree&gt; WORK=&lt;evidence&gt;/work bash drive.sh` - 7 scenes covering --until-event exit codes, config-seconds parsing, status_changed_for dwell durability, promote&#39;s durable state, promote under a live watcher, and both pane-status leak directions</code>
- <code>Before/after control: same driver against a binary built from the parent commit `60b3e0b` (`git archive 60b3e0b` into a temp tree), producing transcript-parent-control.txt</code>
- `Diff of scenes A-E between the parent and head transcripts to confirm the fix changed only scenes F and G`
- <code>Regression-coverage control: HEAD&#39;s `internal/watcher/watcher_test.go` compiled against the parent&#39;s product code - `go test -run &#34;TestForgetPaneScopedCacheClearsEveryPaneAnchoredLatch|TestForgetPaneScopedCacheStopsIdleUnreported|TestForgetPaneScopedCacheLetsTheShipsOwnBlocked&#34; ./internal/watcher/` - fails all three, passes at HEAD</code>

</details>

<details>
<summary>**Document** - 1 issue found -> auto-fixed </summary>

- warning: `internal/watcher/events.go:70` - SPECS.md&#39;s new &#34;Pane-anchored facts across `hand promote`&#34; tables claim to classify every cached fact, and the restart table asserts &#34;Anything added to `TaskState` belongs in this table before it ships&#34;, but `TaskState.Probed` (internal/watcher/events.go:70) appears in neither table. It is the once-only latch behind the `failed &lt;id&gt;` announcement (`ClassifyStatus` fires `failed` only on the true-&gt;false edge), and `forgetPaneScopedCache` (internal/watcher/watcher.go:274) does not reset it. That makes it look pane-anchored-but-carried: a scout whose pane probe failed leaves `Probed` false, and after a promote the ship&#39;s own first probe failure compares as no edge and never announces `failed`. I did not add a table row because either wording would be wrong: documenting it as &#34;genuinely pane-independent, and carried&#34; would enshrine what looks like a real omission, and documenting it as &#34;reset on promote&#34; would describe behavior the code does not have. Resolving it needs a code decision (reset `Probed` in `forgetPaneScopedCache`, or establish why carrying it is correct), which is outside this documentation phase; the table row should follow that decision.

Fix: reset probe latch on promote and document it
Re-checked - no issues remain.

</details>

<details>
<summary>**Lint** - passed</summary>

No issues found.

</details>

<details>
<summary>**Push** - passed</summary>

No issues found.

</details>


atqamz:
@claude re review

PR #225, opened 2026-08-14
refactor(state): make lifecycle transitions atomic

### LINKED ISSUE #194: refactor(state): make lifecycle ownership and persistence atomic
## Parent
Part of #176.

## Goal
Make lifecycle ownership and state changes atomic enough that concurrent commands and restarts cannot create ambiguous ownership.

## Locked authority model

```text
SQLite                         = durable record of intent/history
Git/Treehouse/Herdr/processes = observed evidence of current reality
reconciler                     = deterministic canonical next-state logic
```

SQLite is **not** a distributed transaction with external tools. Persist enough intermediate facts that a crash after any meaningful side effect can later be explained/reconciled.

Prefer atomic SQL conditional mutations over separate read/check/write calls.

Cover contention such as:

```text
spawn vs spawn
send vs send
send vs watcher resume
teardown vs watcher/reconciler
promote vs watcher/send
operator hold vs machine hold
reconciler vs foreground mutation
```

Migrations must be tested from realistic prior schemas.

## Acceptance
- [ ] at most one active attempt is enforced atomically
- [ ] competing writers cannot both acquire lifecycle ownership
- [ ] critical conditional updates are transactional/atomic where possible
- [ ] partial lifecycle facts survive restart
- [ ] SQLite contention/busy behavior is intentional
- [ ] real pre-refactor DB shapes migrate without manual repair
- [ ] no code assumes SQLite atomically includes external tools

## Non-goals
No distributed transaction coordinator, alternate persistence provider, or generic lock service.

## 0.5.0 alignment — atomic Task/Attempt transitions

Depends on #193. P0 before #195/#215.

Keep the existing SQLite strengths: foreign keys, bounded busy timeout, schema migration lock, and transactional `user_version` updates.

### Required atomic invariants

- at most one active (provisioning/running) Attempt per Task;
- terminal Task cannot gain an Attempt except explicit reopen;
- terminal Attempt cannot return to running;
- stale caller cannot overwrite a newer lifecycle transition.

Use conditional SQL transitions (`... WHERE lifecycle = expected`) and a storage-level unique active-attempt constraint. Do not rely only on read/check/write plus optimism.

Portable file/process locks can continue serializing long external critical sections, but the DB transition is the durable truth after a crash.

### Persist external-operation evidence

Provisioning needs durable phase/resource facts around:

```text
Attempt provisioning created
→ worktree lease acquired/recorded
→ Herdr pane acquired/recorded
→ launch submitted
→ launch confirmed
→ Attempt running
```

These are not one transaction. Persist enough after each ownership boundary that #196 can deterministically continue, safely roll back, refuse, or mark repair.

### Narrow mutation API

Replace lifecycle-critical whole-row upserts with narrow methods that transactionally mutate only the invariant they own: create Attempt-if-no-active, record worktree, record pane, mark running, mark terminal, close/reopen Task.

### Additional tests

- concurrent Attempt creation -> exactly one active winner
- active unique index catches bypassed higher-level checks
- stale transition loses instead of overwriting
- terminalization + active relation clear is atomic
- foreground/watcher/reconciler lock ordering cannot deadlock in covered paths
- two concurrent openers safely migrate old DB
- failed migration never advances user_version

atqamz:
## Summary
- Make Task/Attempt ownership transitions and terminalization conditional and transactional inside SQLite.
- Persist provisioning ownership and launch evidence incrementally, with attempt-targeted watcher/send writes and race-safe holds.
- Preserve migration rollback and unstamped split-layout version stamping guarantees.

Fixes atqamz/hand#194

## Test plan
- nix develop --command go test ./... -count=1
- nix develop --command go test -race ./...
- nix develop --command go vet ./...
- nix develop --command make lint
- nix develop --command make build
- nix develop --command make contract
- nix develop --command make e2e
- nix flake check --print-build-logs

### DIFF
--- a/cmd/deliver.go
+++ b/cmd/deliver.go
@@ -47,11 +47,12 @@ func newDeliverCmd() *cobra.Command {
 			// Re-running with a new reason is a correction, not a conflict: nothing
 			// downstream has consumed the mark until teardown reads it, and the last
 			// word on what was delivered is the one worth keeping.
-			t.DeliveredAt = time.Now().UTC().Format(time.RFC3339)
-			t.DeliveredReason = reason
-			if err := state.UpdateTask(home, t); err != nil {
+			deliveredAt := time.Now().UTC().Format(time.RFC3339)
+			if err := state.SetTaskDelivery(home, t.ID, deliveredAt, reason); err != nil {
 				return fmt.Errorf("write task state: %w", err)
 			}
+			t.DeliveredAt = deliveredAt
+			t.DeliveredReason = reason
 
 			var doc axi.Doc
 			doc.Field("id", id)
--- a/cmd/merge.go
+++ b/cmd/merge.go
@@ -125,11 +125,12 @@ func runPRMerge(cmd *cobra.Command, home string, t state.Task, method string) er
 		return fmt.Errorf("gh pr merge failed: %s", strings.TrimSpace(string(out)))
 	}
 
-	t.MergeExecuted = true
-	t.MergeExecutedAt = time.Now().UTC().Format(time.RFC3339)
-	if err := state.UpdateTask(home, t); err != nil {
+	mergedAt := time.Now().UTC().Format(time.RFC3339)
+	if err := state.SetTaskMerge(home, t.ID, mergedAt); err != nil {
 		return fmt.Errorf("write task state: %w", err)
 	}
+	t.MergeExecuted = true
+	t.MergeExecutedAt = mergedAt
 
 	if proj, exists, err := project.Find(home, t.Project); err != nil {
 		return err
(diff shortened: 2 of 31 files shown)


atqamz:
Addressed the teardown launch-evidence blocker in d2414eb.

- Provisioning Attempts with launch-submitted or launch-confirmed evidence now run normal landed-work safety checks.
- Non-forced teardown refuses launched provisioning without a worktree; --force remains explicit.
- Pristine provisioning still records the existing never-launched interrupted outcome.
- Launched provisioning closes as interrupted because provisioning -> completed is not a valid storage transition, while completion evidence no longer claims never launched.
- Added focused regressions and corrected the send lock-order comment.

Local full verification passed; PR CI is running on d2414eb.

Issue #401, opened 2026-08-26
fix(dispatch): mechanical plan staleness check disagrees with itself on the current base sha


atqamz:
## Expected

Dispatching a `mechanical`-class task (`hand spawn`/`hand reopen`) against a project whose
upstream branch has stopped moving should succeed once the brief's `planned_against` matches
the project's current base, and `hand reconcile` on a wedged `provisioning` attempt should
either advance it or report a state that reflects a fresh check.

## Actual

Repeated dispatch attempts for one `mechanical` task (`nsr`, base branch `develop`) kept
failing with a "stale" precondition, each time citing a *different* base sha, even though a
direct check against GitHub (`gh api repos/yes2games/nsr/branches/develop -q .commit.sha`)
showed `develop` had stopped moving and sat at one sha for several minutes:

```
error: "mechanical plan is stale: planned against 66d9255c313a76ac13e60cf79b5b71ee6f8bedb7, current project base is 91b995a4e3357dc4e7cd3d1560ee063cbe9a890b; re-check and rewrite the brief before dispatch"
kind: precondition
exit: 3
```
```
error: "mechanical plan became stale during worktree acquisition: planned against 91b995a4e3357dc4e7cd3d1560ee063cbe9a890b, acquired worktree is 66d9255c313a76ac13e60cf79b5b71ee6f8bedb7; refusing to launch; re-check and rewrite the brief before dispatch"
kind: precondition
exit: 3
```

Note the second error: within *one* command invocation, the "current project base" pre-check
and the worktree-acquisition step disagreed with each other (`91b995a...` vs `66d9255...`),
despite no intervening push - both shas had already been seen in prior/later checks, so this
isn't just "the branch moved mid-command," the two internal checks appear to read from sources
that can disagree.

Editing the brief to chase whichever sha the last error cited never converged - the next
attempt would cite yet another value (a third sha, `7c34f10715cdc1ea9442edbdf397ef5d4917a0f5`,
also appeared across later attempts), including a case where `hand reopen` itself reported
"current project base is 91b995a..." and then, in the very next invocation seconds later,
"acquired worktree is 7c34f107..." with no edit or repo activity in between.

The resulting attempt got permanently wedged in `provisioning` with `worktree: none` for 9+
minutes and never launched. `hand reconcile` on the wedged attempt did not appear to trigger a
fresh check - two separate invocations, several minutes apart, returned byte-identical error
text including an unchanged internal `attempt: 60` counter:

```
count: 1
[...]

PR #404, opened 2026-08-26
fix: scope treehouse pools to fleet homes

### LINKED ISSUE #403: worktree can resolve to a different fleet home's same-named project clone
## Context

Dogfooding rujak work in this fleet home (`/home/atqa/github/atqamz/hand`, `current_home` per
`hand fleet`). Investigating why `no-mistakes runs` showed no history for two in-flight rujak
tasks' own branches, while an older task's branches did show up.

## What I found

Two worktrees for the *same* registered project (`rujak`), spawned in the *same* fleet home,
resolve their git common-dir to two different local repositories:

```
$ git -C <worktree for task rujak-245-release-automation> rev-parse --git-common-dir
/home/atqa/hand-fleet/projects/rujak/.git

$ git -C <worktree for task rujak-250-lineitem-lock-fix> rev-parse --git-common-dir
/home/atqa/github/atqamz/hand/projects/rujak/.git
```

`/home/atqa/hand-fleet` is a separate, independent Secondhand fleet home (its own `data/`,
`state/`, `projects/`, with a materially different project list - `butler`, `hagegames`,
`hageinfra`, `aslab-project`, `nsr-nakama`, `secondhand`, `universe`, `yes2mail` - none of which
this fleet home has ever supervised in this session). It is not listed under any `home` in `hand
fleet`'s registry dump (`registry: /home/atqa/.secondhand/registry.db`) that I could find by
grepping for the literal path.

Despite the worktree pointing at the *other* fleet's clone, every other piece of `rujak-245-
release-automation`'s bookkeeping is correctly rooted in *this* fleet home: its `report_file`,
its `data/rujak-245-release-automation/brief.md`, its entry in `hand status`, all live under
`/home/atqa/github/atqamz/hand/...` as expected. Only the worktree's origin-clone linkage is
wrong.

At least one other task in this fleet's own `data/` (`rujak-235-ci-qualification`, from an
earlier session, predating this one) shows the same pattern - its branches' `no-mistakes` gate
history is attached to the `hand-fleet` clone's gate id, not this fleet's. Two more recent tasks
(`rujak-249-attachment-ux-fix`, `rujak-250-lineitem-lock-fix`, both spawned later in the *same*
session as `rujak-245-release-automation`) correctly resolved to this fleet's own
`projects/rujak` clone. So whatever causes the wrong resolution isn't permanent for the project
- it happened for two tasks and not for two later ones, with no `hand project add`/`hand project
sync` run by me in between that I'm aware of correcting it.

`hand doctor` reports `ready: true`, `violations: 0`, and does not surface this mismatch.

## Impact

- No data loss - the affected tasks' commits are pushed to the real GitHub remote regardless of
  which local clone birthed the worktree, and PRs are open/mergeable normally.
- `no-mistakes` gate history is split across two local-repo identities for what is conceptually
  one project, which is what surfaced this: pushing a branch through the gate from one worktree
  doesn't show up as history when checked from a worktree resolving to the other clone.
- Worse case not observed here but worth naming: if `/home/atqa/hand-fleet` (a fleet home outside
  this session's awareness) is ever cleaned up, garbage-collected, or has its `projects/rujak`
  removed/re-added, a worktree silently depending on it could lose its object source.

## What I'd expect

Either `hand spawn`/`hand project` should always create a task's worktree from *this* fleet
home's own registered clone (never resolve to some other fleet's same-named project), or - if
cross-fleet clone reuse is intentional for some efficiency reason - `hand doctor` should be able
to detect and flag a worktree whose git-common-dir does not point back into the current fleet
home, matching the spirit of the existing "duplicate or identity mismatch" guidance in
`AGENTS.md`/the `secondhand` skill.

I have not touched either fleet home to try to fix this - flagging as found, not attempting a
repair blind to the actual cause.

### LINKED ISSUE #425: TestNonblockingRefusalIsPreservedCrossProcess is flaky: its lock holder uses a bare select{}, so Go's deadlock detector kills it and releases the lock
`TestNonblockingRefusalIsPreservedCrossProcess` in `internal/state/task_test.go` holds a lock in a child process with a bare `select {}`, which Go's runtime deadlock detector kills. When the child dies the OS releases the file lock, and the parent's assertion then sees a free lock.

## Mechanism

The helper the child runs:

```go
func runTaskLockHoldForeverHelper(t *testing.T) {
	_, err := store.Lock(os.Getenv(taskLockHomeEnv), os.Getenv(taskLockKeyEnv), false)
	if err != nil {
		t.Fatalf("holder Lock: %v", err)
	}
	fmt.Println("acquired")
	select {}          // <- line 473
}
```

The child is spawned as `os.Args[0] -test.run=^TestNonblockingRefusalIsPreservedCrossProcess$`, so once it prints `acquired` and enters `select {}` there may be no other runnable goroutine at all. Go then reports

```
fatal error: all goroutines are asleep - deadlock!
```

and the process exits, dropping the lock. The parent, which has already been told `acquired`, asserts the lock is held and finds it is not:

```
--- FAIL: TestNonblockingRefusalIsPreservedCrossProcess (0.06s)
    task_test.go:455: store.Lock got <nil>, want filelock.ErrBusy
FAIL	github.com/atqamz/hand/internal/state	3.499s
```

Whether the detector fires before the parent's assertion runs is pure scheduling luck, so this passes almost always and fails on whichever runner is slowest. Observed on `macos-15-intel` in run 32965684557 on PR atqamz/hand#404, with every other platform green in the same run.

## Why it looks like a lock bug and is not

The failure reads as "non-blocking refusal is broken" - the exact property the test is named for - and the accompanying `deadlock!` line reads as a lock-ordering bug in `internal/filelock` or `internal/store`. Both are misdirection. The lock behaved correctly; the holder stopped holding.

## Fix

Block in the child on something the runtime does not classify as a deadlock. Either works:

```go
for { time.Sleep(time.Hour) }   // a pending timer keeps the runtime from declaring deadlock
```

or read from `os.Stdin`, which also gives the parent a way to release the holder deliberately instead of killing it.

Do not "fix" it by retrying the assertion or by loosening it to accept `nil`; the assertion is the point of the test.

## Note

Not a regression from atqamz/hand#404 - that PR does not touch `internal/state`, `internal/filelock`, or the lock path. Its one `internal/store/store.go` change closes `rows` before running nested queries, which plausibly shifts timing enough to make the existing race lose more often, but the defect is the `select {}`.

atqamz:
## Summary
- Scope Treehouse acquisition, status, return, and initialization to the registered project clone.
- Reject an acquired worktree whose Git common directory differs from the registered clone, and make `hand doctor` report active mismatches.
- Keep mechanical base and acquired-worktree reads on the same repository, addressing the root cause reported in atqamz/hand#401.

Fixes atqamz/hand#403

## Test plan
- `make test`
- `make e2e`
- `make lint`
- `make build`
- Focused foreign-path and same-fleet operator-checkout regressions
- Read-only inspection of the live mismatches; no changes to `/home/atqa/hand-fleet`.

## Summary
- Scope Treehouse acquisition, status, return, and initialization to the registered project clone.
- Reject an acquired worktree whose Git common directory differs from the registered clone, and make `hand doctor` report active mismatches.
- Centralize filesystem path identity in `internal/git`, encode Windows-path fixtures with `encoding/json`, and close task query rows before nested attempt reads to avoid the one-connection SQLite deadlock.
- Load doctor histories at the command boundary, close observation gaps for canonical lease paths, and enforce Herdr composer deadlines.

Fixes atqamz/hand#403

Closes atqamz/hand#425

## Decisions
- Acquisition binds to the registered clone path, not merely the fleet home. `get`, `status`, `return`, and `init` select the pool with the clone path, and acquired worktrees are checked against the registered clone `.git` common directory.
- Slots 3, 4, 5, and 11 are all deferred, not repaired. Slots 3 and 11 alias the `hand4` metadata directory owned by live slot 16; pruning either could touch a live index. Slots 4 and 5 point at missing `hand5` and `hand6` metadata directories, so they are dangling and unusable rather than empty.
- The wider live alias set is tracked in [atqamz/hand#412](https://github.com/atqamz/hand/issues/412): slot 7 aliases `hand` (slot 12), slot 8 aliases `hand1` (slot 13), slot 9 aliases `hand2` (slot 14), slot 10 aliases `hand3` (slot 15), and slots 3 and 11 alias `hand4` (slot 16). Repair waits until 403, 406, 407, and 408 have landed because several aliased metadata directories belong to live workers.
- Registered-clone binding is necessary but not sufficient for those aliases: both sides resolve to the correct clone common directory, so this PR deliberately does not try to detect them. Slots 1, 2, and 6 instead point at the fleet home repository and remain distinct wrong-repository findings.

## Test plan
- `make test` (race-enabled full suite)
- `make e2e`
- `make lint`
- `make build`
- Focused foreign-path, same-fleet operator-checkout, canonical path, and Windows-safe JSON regressions
- Read-only inspection of live mismatches; no changes to `/home/atqa/hand-fleet`
- Windows test run: `cmd` 453.6s and `internal/runtime` 330.8s, both below the 600s per-package cap; main baseline is 353.6s and 217.0s, with additional headroom tracked in [atqamz/hand#416](https://github.com/atqamz/hand/issues/416). The timeout cap was not raised.

### DIFF
--- a/cmd/doctor.go
+++ b/cmd/doctor.go
@@ -4,11 +4,13 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"runtime"
 	"strconv"
 	"strings"
 
 	"github.com/atqamz/hand/internal/agentsmd"
 	"github.com/atqamz/hand/internal/axi"
+	"github.com/atqamz/hand/internal/git"
 	"github.com/atqamz/hand/internal/harness"
 	"github.com/atqamz/hand/internal/home"
 	"github.com/atqamz/hand/internal/integration"
@@ -250,6 +252,7 @@ func doctorFindings(fleetHome string, histories []state.TaskHistory) ([]doctorFi
 			findings = append(findings, doctorFinding{Severity: doctorError, Text: fmt.Sprintf("project %q no-mistakes gate is %s", p.Name, issue)})
 		}
 	}
+	findings = append(findings, doctorWorktreeFindings(fleetHome, histories)...)
 
 	detection, err := harness.DetectCurrent()
 	if err != nil {
@@ -322,6 +325,50 @@ func reportPathSuffix(text string, end int) bool {
 
 func isReportPathSuffixChar(char byte) bool {
 	return (char >= 'a' && char <= 'z') || (char >= 'A' && char <= 'Z') || (char >= '0' && char <= '9') || char == '_' || char == '-' || char == '/' || char == '\\'
+}
+
+func doctorWorktreeFindings(fleetHome string, histories []state.TaskHistory) []doctorFinding {
+	findings := make([]doctorFinding, 0)
+	for _, history := range histories {
+		attempt := history.ActiveAttempt
+		if attempt == nil || attempt.Worktree == "" {
+			continue
+		}
+		clone := filepath.Join(fleetHome, "projects", history.Task.Project)
+		expected := filepath.Join(clone, ".git")
+		if _, err := os.Stat(expected); err != nil {
+			findings = append(findings, doctorFinding{Severity: doctorError, Text: fmt.Sprintf("task %q registered clone cannot be inspected: %v", history.Task.ID, err)})
+			continue
+		}
+		actual, err := git.CommonDir(attempt.Worktree)
+		if err != nil {
+			findings = append(findings, doctorFinding{Severity: doctorError, Text: fmt.Sprintf("task %q worktree common directory cannot be inspected: %v", history.Task.ID, err)})
+			continue
+		}
+		if sameDoctorPath(expected, actual) {
+			continue
+		}
+		findings = append(findings, doctorFinding{Severity: doctorError, Text: fmt.Sprintf("task %q worktree is rooted in another Git repository: got %s, want %s", history.Task.ID, actual, expected)})
+	}
+	return findings
+}
+
+func sameDoctorPath(left, right string) bool {
+	canonical := func(path string) string {
+		abs, err := filepath.Abs(path)
+		if err != nil {
+			return filepath.Clean(path)
+		}
+		if resolved, err := filepath.EvalSymlinks(abs); err == nil {
+			abs = resolved
+		}
+		return filepath.Clean(abs)
+	}
+	left, right = canonical(left), canonical(right)
+	if runtime.GOOS == "windows" {
+		return strings.EqualFold(left, right)
+	}
+	return left == right
 }
 
 // A local-only fleet never needs gh, while a registered project delivering through direct-pr or
--- a/cmd/doctor.go
+++ b/cmd/doctor.go
@@ -4,7 +4,6 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
-	"runtime"
 	"strconv"
 	"strings"
 
@@ -345,30 +344,12 @@ func doctorWorktreeFindings(fleetHome string, histories []state.TaskHistory) []d
 			findings = append(findings, doctorFinding{Severity: doctorError, Text: fmt.Sprintf("task %q worktree common directory cannot be inspected: %v", history.Task.ID, err)})
 			continue
 		}
-		if sameDoctorPath(expected, actual) {
+		if git.SamePath(expected, actual) {
 			continue
 		}
 		findings = append(findings, doctorFinding{Severity: doctorError, Text: fmt.Sprintf("task %q worktree is rooted in another Git repository: got %s, want %s", history.Task.ID, actual, expected)})
 	}
 	return findings
-}
-
-func sameDoctorPath(left, right string) bool {
-	canonical := func(path string) string {
-		abs, err := filepath.Abs(path)
-		if err != nil {
-			return filepath.Clean(path)
-		}
-		if resolved, err := filepath.EvalSymlinks(abs); err == nil {
-			abs = resolved
-		}
-		return filepath.Clean(abs)
-	}
-	left, right = canonical(left), canonical(right)
-	if runtime.GOOS == "windows" {
-		return strings.EqualFold(left, right)
-	}
-	return left == right
 }
 
 // A local-only fleet never needs gh, while a registered project delivering through direct-pr or
--- a/cmd/doctor_test.go
+++ b/cmd/doctor_test.go
@@ -226,6 +226,90 @@ func hasDoctorFindingPrefix(findings []doctorFinding, severity doctorSeverity, p
 	return false
 }
 
+func TestDoctorFindsWorktreeUsingAnotherFleetClone(t *testing.T) {
+	home := t.TempDir()
+	t.Chdir(home)
+	mkFleetDirs(t, home)
+	if _, err := agentsmd.Refresh(home); err != nil {
+		t.Fatal(err)
+	}
+	if _, err := skill.Refresh(home); err != nil {
+		t.Fatal(err)
+	}
+	mustConfigSet(t, settingHarness, harness.Claude)
+	if err := project.Add(home, project.Project{Name: "demo", URL: "https://example.com/demo.git", Mode: project.ModeLocalOnly}); err != nil {
+		t.Fatal(err)
+	}
+	clone := filepath.Join(home, "projects", "demo")
+	foreign := filepath.Join(t.TempDir(), "demo")
+	initGitRepo(t, clone)
+	initGitRepo(t, foreign)
+	if err := state.CreateTask(home, state.Task{ID: "task-1", Project: "demo"}); err != nil {
+		t.Fatal(err)
+	}
+	if _, err := state.CreateAttempt(home, state.Attempt{TaskID: "task-1", Lifecycle: state.AttemptRunning, Worktree: foreign}); err != nil {
+		t.Fatal(err)
+	}
+	histories, err := state.ListOpenHistoriesReadOnly(home)
+	if err != nil {
+		t.Fatal(err)
+	}
+	findings, err := doctorFindings(home, histories)
+	if err != nil {
+		t.Fatal(err)
+	}
+	found := false
+	for _, finding := range findings {
+		if finding.Severity == doctorError && strings.Contains(finding.Text, `task "task-1" worktree is rooted in another Git repository`) {
+			found = true
+			break
+		}
+	}
+	if !found {
+		t.Fatalf("findings = %#v, want a cross-fleet worktree finding", findings)
+	}
+}
+
+func TestDoctorFindsWorktreeUsingTheFleetHomeCheckout(t *testing.T) {
+	home := t.TempDir()
+	t.Chdir(home)
+	mkFleetDirs(t, home)
+	if _, err := agentsmd.Refresh(home); err != nil {
+		t.Fatal(err)
+	}
+	if _, err := skill.Refresh(home); err != nil {
+		t.Fatal(err)
+	}
+	mustConfigSet(t, settingHarness, harness.Claude)
+	if err := project.Add(home, project.Project{Name: "demo", URL: "https://example.com/demo.git", Mode: project.ModeLocalOnly}); err != nil {
+		t.Fatal(err)
+	}
+	clone := filepath.Join(home, "projects", "demo")
+	operatorCheckout := filepath.Join(home, "operator-checkout")
+	initGitRepo(t, clone)
+	initGitRepo(t, operatorCheckout)
+	if err := state.CreateTask(home, state.Task{ID: "task-1", Project: "demo"}); err != nil {
+		t.Fatal(err)
+	}
+	if _, err := state.CreateAttempt(home, state.Attempt{TaskID: "task-1", Lifecycle: state.AttemptRunning, Worktree: operatorCheckout}); err != nil {
+		t.Fatal(err)
+	}
+	histories, err := state.ListOpenHistoriesReadOnly(home)
+	if err != nil {
+		t.Fatal(err)
+	}
+	findings, err := doctorFindings(home, histories)
+	if err != nil {
+		t.Fatal(err)
+	}
+	for _, finding := range findings {
+		if finding.Severity == doctorError && strings.Contains(finding.Text, `task "task-1" worktree is rooted in another Git repository`) {
+			return
+		}
+	}
+	t.Fatalf("findings = %#v, want an operator-checkout worktree finding", findings)
+}
+
 func TestDoctorIncludesProjectListGateFinding(t *testing.T) {
 	home := t.TempDir()
 	t.Chdir(home)
--- a/cmd/project.go
+++ b/cmd/project.go
@@ -628,7 +628,7 @@ func treehouseInitIfNeeded(clonePath string) error {
 	if _, err := os.Stat(filepath.Join(clonePath, "treehouse.toml")); err == nil {
 		return excludeLocally(clonePath, "treehouse.toml")
 	}
-	out, err := runManagedCore(context.Background(), "treehouse", clonePath, "init")
+	out, err := runManagedCore(context.Background(), "treehouse", clonePath, "--root", clonePath, "init")
 	if err != nil {
 		return fmt.Errorf("treehouse init failed: %s", string(out))
 	}
--- a/cmd/promote_test.go
+++ b/cmd/promote_test.go
@@ -482,7 +482,7 @@ func TestPromoteReturnsScoutResourcesWhenTheShipLaunchFails(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
-	if !strings.Contains(string(calls), "treehouse return "+oldWt) {
+	if !strings.Contains(string(calls), " return "+oldWt) {
 		t.Fatalf("calls = %q, want the scout worktree returned", calls)
 	}
 	if !strings.Contains(string(calls), "tab close wA:tOld") {
--- a/cmd/reopen_test.go
+++ b/cmd/reopen_test.go
@@ -1,6 +1,8 @@
 package cmd
 
 import (
+	"os/exec"
+	"path/filepath"
 	"strings"
 	"testing"
 
@@ -18,7 +20,12 @@ func TestReopenCreatesANewAttemptWithoutResurrectingTheOldOne(t *testing.T) {
 	}
 	worktree := t.TempDir() + "/wt"
 	home := setupSpawnHome(t, worktree, herdr)
-	initGitRepo(t, worktree)
+	clone := filepath.Join(home, "projects", "myproj")
+	initGitRepo(t, clone)
+	command := exec.Command("git", "-C", clone, "worktree", "add", "--detach", worktree)
+	if output, err := command.CombinedOutput(); err != nil {
+		t.Fatalf("git worktree add: %v: %s", err, output)
+	}
 
 	spawn := newSpawnCmd()
 	spawn.SetArgs([]string{"task-1", "myproj", "--harness", harness.Claude})
--- a/cmd/teardown_test.go
+++ b/cmd/teardown_test.go
@@ -1555,8 +1555,11 @@ func readTreehouseReturnArgs(t *testing.T, worktree string) []string {
 	t.Helper()
 	var args []string
 	for _, line := range readInvocations(t, worktree) {
-		if fields := strings.Fields(line); len(fields) > 1 && fields[1] == "return" {
-			args = fields[1:]
+		fields := strings.Fields(line)
+		for i, field := range fields {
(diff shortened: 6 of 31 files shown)


atqamz:
Supervisor review. All three non-Linux jobs have finished and all three fail. Full evidence here so
it does not have to be rediscovered. Two of these belong to this PR; the third probably does not.

**1. `TestCommonDirReturnsTheMainRepositoryForAWorktree` - macos-latest, macos-15-intel AND
windows-latest.** The two spellings differ in an instructive way:

```
windows: CommonDir() = "C:\Users\runneradmin\AppData\Local\Temp\...\repo\.git"
                 want "C:\Users\RUNNER~1\AppData\Local\Temp\...\repo\.git"
macos:   CommonDir() = "/private/var/folders/.../repo/.git"
                 want "/var/folders/.../repo/.git"
```

macOS is the `/var` -> `/private/var` symlink. Windows is 8.3 short-name expansion: `TMP` yields
`RUNNER~1` while git answers `runneradmin`. Both are the same fact stated twice - one directory has
several legal spellings, and this fix compares path strings. That makes it a production defect, not
a test artifact: this change binds a Treehouse slot to the project's registered clone and makes
`hand doctor` flag any slot whose common-dir is something else, and both are path comparisons. If
one side is canonical and the other is not, then on macOS and Windows every healthy slot compares
unequal - doctor reports false mismatches and acquisition refuses valid worktrees. Failing closed on
correct state is worse than the bug being fixed, and both platforms are supported targets.

Go's `filepath.EvalSymlinks` covers both cases (on Windows it expands 8.3 to the long form), so one
canonicalize helper - EvalSymlinks then Clean - is enough, applied in production code at the
boundary where a filesystem path becomes a comparable identity, for the stored/derived registered
clone path and the git-reported common-dir alike. Decide deliberately what happens when
EvalSymlinks errors on a path that does not exist yet; do not let that fall through to an
unresolved comparison. Do not fix this by relaxing the test's expectation, and do not resolve
symlinks only inside the test.

**2. `TestObserveLeaseScopesStatusToTheFleetHome` - windows-latest.** A genuine bug in this PR, not a
platform quirk:

```
worktree_test.go:389: ... Reason:treehouse status output is not a JSON array:
                      invalid character 'U' in string escape code
```

JSON is being built by interpolating a path into a string. On Windows the path contains
backslashes, so `C:\Users\...` produces the invalid escape `\U` and the document does not parse.
Marshal properly (`json.Marshal` over a struct or map) instead of concatenating. Worth checking
whether this shape exists only in `internal/faketool/treehouse.go` or in production code too - if
production builds JSON this way, that is a separate real defect.

**3. `TestNonblockingRefusalIsPreservedCrossProcess` - macos-15-intel only.** Probably not this PR's:

```
task_test.go:455: store.Lock got <nil>, want filelock.ErrBusy
```

This diff does not touch `internal/state`, and `main` is green on all six of its most recent runs
including that job. It reads like the load-sensitive cross-process lock flake this repo has hit
before. Confirm by re-running that one job; if it goes green, say so and leave it alone. Do not
weaken the assertion or add a sleep - if it proves reproducible it is its own issue, not part of
#403.

Also still needed in the PR body, and not answered yet:

- Does acquisition bind a slot to the **project's registered clone path**, or only to the fleet
  home? Fleet-home scoping alone is insufficient. The worst live instance -
  `/home/atqa/.treehouse/hand-a7a4d5/6/hand`, whose common-dir is
  `/home/atqa/github/atqamz/hand/.git` - points at the operator's own source checkout, which is
  *inside* the correct fleet home. The `hand doctor` check described here uses the right invariant
  ("outside the registered clone"); it needs to be enforced at acquisition too, not only reported
  afterwards.
- Are slots 3/4/5/11 covered? They return an empty common-dir, and 4 and 11 both claim worktree
  entry `hand4`. Repaired, named by doctor, or deferred to their own issue - pick one and say which.

atqamz/hand#401 is formally held as blocked on this, because its repro is this bug: the "current
project base" read and the "acquired worktree" read resolve to two different git dirs, so no
rewrite of `planned_against` can make them agree. If this change makes #401 stop reproducing, say so
in the body, but do not add a closing keyword for it - that call comes after a re-test.

atqamz:
Separate from the review above, and my fault rather than this PR's: this task has never emitted a
single `hand` report, because its brief never declared the report channel. `internal/state/report.go`
states the contract - the `.status` file is a channel the worker appends to via plain shell
redirection, and "its absolute path belongs in the brief the supervisor writes - hand only ever reads
it, never writes it." Without that line there is no way to know where to write. Filed the structural
gap as atqamz/hand#406.

For the rest of this task, append each report as one line to:

```
/home/atqa/github/atqamz/hand/state/403-cross-fleet-worktree-resolution.status
```

using the report vocabulary: `working:`, `done:`, `failed:`, `blocked:`, `needs-decision:`,
`paused:`. Please report where you currently are, then keep reporting as you go - I have been reading
GitHub instead of hand for this whole task.

(Tried to send this through the composer twice; both attempts refused cleanly with
`send-not-submitted` / `composer-timeout` because the agent is genuinely busy, so this comment is the
durable copy.)

atqamz:
Supervisor review of PR #404 head 0fa0728289.

macOS FAILURE is yours and it is a test bug, not a production bug. internal/git/repository_test.go:41 TestCommonDirReturnsTheMainRepositoryForAWorktree asserts got == filepath.Join(repo, ".git") where repo comes from a raw t.TempDir(). On macOS t.TempDir() hands back /var/folders/... which is a symlink to /private/var/folders/..., and git rev-parse --git-common-dir answers with the resolved form. CommonDir deliberately does not EvalSymlinks, so that assertion can never hold there. Fix the test, not CommonDir: canonicalize both sides before comparing, the same way sameDoctorPath does (filepath.Abs, then filepath.EvalSymlinks, ignoring the EvalSymlinks error and keeping the abs path).

While you are there: sameDoctorPath lives in package cmd, but path identity is a git concern that #401 also needs. Move that canonicalize-and-compare into internal/git as an exported helper and have cmd/doctor.go and the new test both call it, so there is exactly one definition of "same path" in the tree. Keep the Windows strings.EqualFold branch.

Windows Test was still pending when I looked; check it after you push.

Two questions in your own PR body are still unanswered and I need both before I can merge: (1) does acquisition bind to the registered clone path or only to the fleet home, and (2) slots 3/4/5/11 have empty common-dirs and 4 and 11 both claim worktree entry hand4 - repaired, reported, or deliberately deferred.

Report channel: append every state change to /home/atqa/github/atqamz/hand/state/403-cross-fleet-worktree-resolution.status with plain shell redirection, e.g.
  echo "working: fixing macOS path assertion" >> /home/atqa/github/atqamz/hand/state/403-cross-fleet-worktree-resolution.status
Vocabulary: working:, done:, failed:, blocked:, needs-decision:, paused:. Nothing you say in the pane reaches me; that file is the only channel. done: never means merged - merge and close are the supervisor's action only, however confident you are that it is ready.

atqamz:
Supervisor, follow-up on PR #404. Both your Decisions answers are what I needed - "acquisition binds to the registered clone path, not merely the fleet home" is the invariant, and naming each of slots 3/4/5/11 individually is right. Your row-close fix in internal/store/store.go is the most valuable thing in the PR: closing the parent rows before the nested ListAttempts read, in both ListOpenTaskHistories and ListReconciliationHistories, fixes a latent one-connection SQLite deadlock that had nothing to do with this task. Good catch on the sibling function nobody asked about.

One correction to your slot findings, and one change I want in the PR body.

Slots 4 and 5 are not "uninspectable empty common-dir". Their .git files name worktrees/hand5 and worktrees/hand6, and those metadata directories do not exist, which is why rev-parse fails with "fatal: not a git repository: (null)". That is a dangling pointer, not an empty one, and it is worth saying precisely.

The duplicate hand4 claim is much worse than a duplicate. I mapped every slot's gitdir target against the clone's metadata directories: the pool has 13 slot directories and the clone has exactly 5 metadata directories, so EVERY free slot aliases a leased one. hand<-7,12; hand1<-8,13; hand2<-9,14; hand3<-10,15; hand4<-3,11,16. Slots 13, 14, 15 and 16 are live workers right now, and slots 8, 9, 10, 3 and 11 are all unleased in treehouse-state.json. So the next hand spawn gets a directory sharing HEAD, index and index.lock with a running worker. Your own slot 13 is aliased by free slot 8. I filed it as atqamz/hand#412 with the full measurement.

What I want from you: do not expand this PR to fix #412 - it is a separate defect with its own issue. Instead replace "deliberately deferred to fleet reconciliation" in your Decisions section with an explicit reference to atqamz/hand#412, so the deferral is tracked rather than just stated, and add one line noting that your registered-clone binding is necessary but not sufficient against it, because both sides of every alias report the correct common directory and therefore pass your new doctor check.

Everything else stands. Report when CI is green on all five platforms. Do not merge; that is mine.

atqamz:
Windows timeout: I measured the thing you need and it changes the diagnosis. Read this before you touch it again.

The panic dump is misleading. It names `TestTeardownRetriesAfterReportRemovalFails (1s)` - that test had been running for one second when the 10-minute alarm fired. It is not hung. There is no deadlock in this run: the only two live goroutines are that test's, blocked in a Windows `FlushFileBuffers` syscall (a journal fsync inside `store.setMeta` under `project.Migrate`), and `database/sql`'s connectionOpener sitting in select. The alarm simply caught a random test. The whole `cmd` package ran out of budget.

Here is the number that matters. On `main`, the same job, same runner:

    ok  github.com/atqamz/hand/cmd  353.626s     (windows-latest, run 32952412368)

353s of a 600s cap. `main` already spends 59% of the Windows timeout in `cmd` with 246s of headroom, and your change spends the rest. macOS-intel runs the same package in 182s and ubuntu in 144s, so Windows is 2.5x slower per test and it is fsync-bound, not CPU-bound - every state-database write pays a real `FlushFileBuffers`.

So stop looking for an unclosed handle or a lock. The mechanism is: `doctorWorktreeFindings` calls `state.ListOpenHistoriesReadOnly(fleetHome)` on **every** `doctor` invocation, that open runs the registry migration path, and the migration write costs an fsync. Multiply by every doctor test in `cmd` plus your three new ones and the package crosses 600s.

Two things to do, in this order.

1. **Do not open the state database inside `doctorFindings`.** Take already-loaded state from the caller, or make the worktree check lazy so a `doctor` run with no open tasks recorded never touches the database at all. Most doctor tests have no tasks; they should pay nothing. This is the fix, and it is the one your own PR body already floated as the right shape.

2. **Report the headroom, do not consume it.** `cmd` at 353s on `main` is a latent CI fragility that predates you: any PR adding a handful of database-touching tests to `cmd` fails with a timeout that reads as a deadlock, which is exactly the hour this cost us. Do not attempt to fix that in this PR and do not raise the timeout. Say in your PR body what `main`'s baseline is and that it is separately tracked, and I will carry it as its own issue.

`internal/runtime` is fixed - it was timing out at 600s before and is under the cap now. `internal/worktree` went 16.9s to 36.7s on Windows from the new fixtures, which is fine.

Still owed in the PR body, from my earlier message: the atqamz/hand#412 reference in place of "deliberately deferred to fleet reconciliation", the corrected slots 4/5 description (dangling `hand5`/`hand6`, not "empty"), and the line that registered-clone binding is necessary but not sufficient.

Do not merge and do not close the issue.

atqamz:
One coordination point that is not visible from inside this PR.

PR atqamz/hand#409 is also adding a state-database read to `doctorFindings` - `state.ListOpen(fleetHome)`, for a "brief does not declare report path" warning - plus three doctor tests of its own. You are adding `state.ListOpenHistoriesReadOnly(fleetHome)` and three doctor tests. Measured on the same `windows-latest` runner:

| branch | `cmd` |
| --- | --- |
| `main` (run 32952412368) | 353.6s |
| atqamz/hand#409 (run 32955808867) | 380.8s |
| this PR (run 32955191776) | timeout at 600s |

Neither of you is doing anything unreasonable. You are independently spending one 246-second budget, and whichever merges second inherits the other's cost - so this PR going green alone is not proof it stays green after atqamz/hand#409 lands, or the reverse.

`doctorFindings` needs to open the store **once** and pass what it read to both checks. You are already reading task histories; atqamz/hand#409 needs the open tasks, which are available from `.Task` on the histories you have. Leave that seam explicitly: read once near the top of `doctorFindings`, hand the result to your worktree check, and shape it so atqamz/hand#409's check can take the same value rather than opening again. Say in your PR body what you left and what its consumer is called - I am giving atqamz/hand#409's worker the mirror of this note, and the two of you must not both write it.

Also note atqamz/hand#409 uses the migrating `state.ListOpen` rather than a read-only variant. Yours is already read-only; keep it that way, and if you add the shared accessor, make it the read-only one. `hand doctor` must not migrate anything.

The 600s cap itself is filed as atqamz/hand#416. Do not raise it here.

atqamz:
Answering open question 2 from your brief myself, so it stops blocking this PR. Full evidence and the method are in atqamz/hand#412 (issuecomment-5424218457); I enumerated all sixteen slots by reading `.git` and worktree-metadata files with `cat`, never invoking git inside a slot that aliases a live worker.

**The answer for every slot is: deferred, not repaired, and deliberately so.** Put that in the PR body with this reasoning:

Six stale slots alias the five worktree metadata dirs that the current generation of workers owns - slot 7 aliases `hand` (slot 12), 8 aliases `hand1` (slot 13, this task), 9 aliases `hand2` (slot 14), 10 aliases `hand3` (slot 15), and slots 3 *and* 11 both alias `hand4` (slot 16). Four of those five dirs belong to workers running right now, yours included. Pruning or repairing any of them touches a live index. Repair waits until 403, 406, 407 and 408 have all landed.

Slots 4 and 5 point at `hand5` and `hand6`, which do not exist. Those two are the only genuinely free slots in the pool, and they are unusable, which is why `hand spawn` and `hand reopen` are frozen for project `hand`.

Slots 1, 2 and 6 are worktrees of `/home/atqa/github/atqamz/hand/.git` - the fleet home's own repository - rather than `projects/hand/.git`. Slot 6 is task `401-mechanical-staleness-fix`. This is the exact defect this PR is about, observed live in this pool, and it is worth citing in the PR body as the motivating case for `doctorWorktreeFindings`: your check compares a task's worktree common-dir against `projects/<project>/.git`, and against this pool it would flag three slots immediately.

Two consequences for your change, both worth a sentence in the PR body:

1. `doctorWorktreeFindings` has to report the wrong-repository case (slots 1, 2, 6) and the missing-metadata case (slots 4, 5) distinctly - "common-dir is not the registered clone" and "common-dir cannot be determined" are different operator actions, and right now the second one renders as an empty string.
2. The aliasing case (slots 3, 7-11) is invisible to a common-dir comparison, because an aliased slot resolves to the *correct* clone. Do not try to catch it here. Say in the PR body that it is out of scope and tracked in atqamz/hand#412, so nobody later reads this check as covering it.

Nothing else in the brief changes. Do not repair any slot yourself.

atqamz:
PR #409 now leaves the shared doctor seam: newDoctorCmd calls ListOpenHistoriesReadOnly once and passes the resulting []state.TaskHistory into doctorFindings. The report-path check consumes history.Task, so the worktree check can consume the same histories without another database open or migration.

atqamz:
Windows is green, and the numbers are better than "the timeout is gone". Recording them because the fix turns out to help `main`, not just this PR.

`windows-latest`, per-package, against Go's 600s default cap:

| package | `main` before | this PR, first attempt | this PR now |
| --- | --- | --- | --- |
| `cmd` | 353.6s | 453.6s | **357.9s** |
| `internal/runtime` | 217.0s | 330.8s | **249.3s** |
| `internal/watcher` | 289.1s | - | **109.3s** |

Two things follow.

**Your `internal/store/store.go` change is the fix, and it is a good one.** Closing `rows` in `ListOpenTaskHistories` and `ListReconciliationHistories` before the loop that runs `db.ListAttempts` per row stops the outer cursor holding a connection while nested queries run. That took `cmd` from +100s over main to +4s, and `internal/runtime` from +114s to +32s.

**`internal/watcher` dropped 62% below main** - 289.1s to 109.3s - and this PR does not touch `internal/watcher` at all. So the connection hogging was a pre-existing cost on `main` that nothing had attributed, and you have removed it. That is worth a line in the PR body: the change is a performance fix for the whole Windows suite, not only a fix for the timeout this PR introduced.

It also retires the goroutine-dump theory for good. The earlier dump showed a single goroutine inside `modernc.org/sqlite` under `database/sql` and read exactly like a lock-ordering deadlock. It was connection starvation plus Windows fsync cost, which is why raising the timeout would have hidden it and why collecting more stacks was never going to find it.

Remaining: `macos-15-intel` only, on `TestNonblockingRefusalIsPreservedCrossProcess`, which is not yours - see atqamz/hand#425.

PR #409, opened 2026-08-26
wire worker report channel into launches

### LINKED ISSUE #406: nothing enforces that a brief declares the report channel, so a worker silently never reports
## What this is

Root cause behind most of what atqamz/hand#402 observed. Filing it separately because #402's other
half - the `state: done` / `attempt_lifecycle: running` contradiction that never converges - is a
genuinely different defect with its own fix in #405.

## The contract

`internal/state/report.go` states it exactly:

> ReportPath returns the path to a task's reported-state channel. The worker appends to it via plain
> shell redirection (**its absolute path belongs in the brief the supervisor writes**) - hand only
> ever reads it, never writes it.

And `skills/secondhand/references/planning-and-briefs.md:78` requires the brief to carry "the
absolute path to `state/<id>.status` and the report vocabulary."

So a worker that was never told the path cannot report, and hand is behaving correctly when it says
`reported: none`. The report channel is not broken; it was never wired.

## The problem

Nothing enforces or supplies it.

- `internal/brief/brief.go` validates `planned_against` (and rejects a non-hex value) but has no
  notion of a report path at all.
- `report_file` exists only as a display column in `cmd/statusview.go`. It is computed from the task
  id at read time, never checked against the brief.
- `hand spawn` accepts a brief with no report path and launches happily. `hand doctor` says nothing.
- hand already rewrites brief front matter at launch - `internal/harness/harness.go:292` appends
  "Any model, effort, execution_class, or planned_against keys in its leading '---' block are
  dispatch metadata, not task instructions" - so the machinery to inject the path is already in the
  launch path.

The failure is silent in the worst way: the worker does real, good work, the supervisor sees
`reported: none` / `report_history[0]`, and the natural reading is "hand's report pipeline is
broken." That is what #402 concluded, and it is wrong.

## Evidence

Every brief under `data/` in this fleet home, checked for a `report_file:` line, against whether
that task ever emitted a single report. 13 for 13, no exceptions:

| brief carries `report_file:` | tasks | reports emitted |
|---|---|---|
| no  | 354-native-runner-lanes, 23-launcher-activation, 24-font-wallpaper-ownership, 39-tray-sni-proof, 6-adoption-docs, 6-consumer-fixture, 401-mechanical-staleness-fix, 402-report-pipeline-fix, 403-cross-fleet-worktree-resolution | 0 for all 9 |
| yes | butler-broker-allowlist-rujak-release, rujak-245-release-automation, cf-access-service-token-rujak-health, rujak-250-lineitem-lock-fix | 4 entries each |

Worth recording because it is a trap: the first correlation this looked like was harness-shaped -
8 of 8 `codex` tasks mute, 2 of 3 `claude` tasks reporting - which would have sent a fix at the
codex report path. It was a confound. The one `claude` task that never reported
(`354-native-runner-lanes`) also has no `report_file:` line, and the four briefs that do carry it
span both harnesses and four different projects. Harness is irrelevant; the brief field is
everything.

Note that #402's own brief mentions `state/<id>.status` twice - but only as prose describing the
bug, not as an instruction with an absolute path. Mentioning the channel is not declaring it.

## Fix direction

Do not fix this only in the skill's prose - the prose already says to do it, and it still did not
happen nine times. Make it structural. Options, roughly in order of preference:

1. **Inject it at launch.** hand knows the home and the task id, so it can compute the absolute path
   and state it in the launch prompt unconditionally, next to where
   `internal/harness/harness.go:292` already annotates front matter. Then a brief cannot omit it,
   and old briefs start working with no rewrite.
2. **Validate it at spawn.** Teach `internal/brief/brief.go` a `report_file` field and have
   `hand spawn`/`hand reopen` refuse, or at minimum warn loudly, when it is absent or does not match
   the task's own `state/<id>.status`. Stronger guarantee, but it invalidates every existing brief.
3. **Surface it as drift.** Have `hand doctor` report live tasks whose brief does not declare the
   channel. Weakest - it only tells you after the fact - but cheap, and complements either of the
   above.

Option 1 plus option 3 gets both "cannot happen again" and "tell me about the ones already broken".

## Acceptance

- A task spawned from a brief with no report path still emits reports.
- The 9 currently-mute tasks are either fixed in place or explicitly named as unfixable, rather than
  quietly staying mute.
- `reported: none` on a task whose worker is doing real work is no longer reachable without hand
  saying why.

Related: atqamz/hand#402

atqamz:
## Summary
- Inject each task absolute `state/<id>.status` path and the six report prefixes into every prompt-capable worker launch.
- The shared provisioning path covers `hand spawn`, `hand reopen`, and `hand promote`; `reconcile.go` also rebuilds the same absolute path for persisted launch confirmation, so reopen preserves the injected path.
- `hand doctor` opens open task histories once through `ListOpenHistoriesReadOnly` and passes the histories into its checks, leaving the seam for atqamz/hand#404's worktree check to reuse without another database open or migration.
- Doctor reports unreadable briefs as task-specific findings and continues scanning, and compares normalized absolute report paths.
- Empty report paths fail prompt-capable harness builds instead of emitting a false instruction.
- Preserve existing briefs and do not add a `report_file` front-matter key.
- Scope: prompt injection covers four of six harnesses. `grok` and `pi` pass the brief as a file with no prompt, so their workers remain without the report channel; follow-up atqamz/hand#418 tracks that limitation.

Before, a constructed Claude prompt for a brief with no report path was:
`Read the brief at /tmp/data/task-1/brief.md and carry out the task it describes.`

After, the same constructed prompt is:
`Read the brief at /tmp/data/task-1/brief.md and carry out the task it describes. The worker report channel is /tmp/state/task-1.status. Append every state change to that file with plain shell redirection; this is the only way anything you say reaches the supervisor. Use these report prefixes: working:, done:, failed:, blocked:, needs-decision:, paused:.`

Closes atqamz/hand#406

Related: atqamz/hand#418 - the `grok` and `pi` gap is tracked there and only there, so #406 is not held open duplicating it.

## Test plan
- `make test` - pass, including the full race suite.
- `make lint` - pass with `0 issues` and commentlint complete.
- Focused harness, doctor, and runtime path tests - pass.
- CI matrix covers Ubuntu x86/ARM, macOS ARM/Intel, and Windows; build and bootstrap E2E jobs remain enabled.

### DIFF
--- a/cmd/doctor.go
+++ b/cmd/doctor.go
@@ -2,6 +2,7 @@ package cmd
 
 import (
 	"fmt"
+	"os"
 	"path/filepath"
 	"strconv"
 	"strings"
@@ -15,6 +16,7 @@ import (
 	"github.com/atqamz/hand/internal/routing"
 	"github.com/atqamz/hand/internal/selfupdate"
 	"github.com/atqamz/hand/internal/skill"
+	"github.com/atqamz/hand/internal/state"
 	"github.com/atqamz/hand/internal/supervision"
 	"github.com/atqamz/hand/internal/toolchain"
 	"github.com/spf13/cobra"
@@ -214,6 +216,21 @@ func doctorFindings(fleetHome string) ([]doctorFinding, error) {
 			if !onPath(tool) {
 				findings = append(findings, doctorFinding{Severity: doctorWarning, Text: fmt.Sprintf("required tool %q is not on PATH", tool)})
 			}
+		}
+	}
+
+	tasks, err := state.ListOpen(fleetHome)
+	if err != nil {
+		return nil, err
+	}
+	for _, task := range tasks {
+		briefPath := filepath.Join(fleetHome, task.Brief)
+		data, err := os.ReadFile(briefPath)
+		if err != nil && !os.IsNotExist(err) {
+			return nil, fmt.Errorf("read task %q brief: %w", task.ID, err)
+		}
+		if !strings.Contains(string(data), state.ReportPath(fleetHome, task.ID)) {
+			findings = append(findings, doctorFinding{Severity: doctorWarning, Text: fmt.Sprintf("task %q brief does not declare report path", task.ID)})
 		}
 	}
 
--- a/cmd/doctor.go
+++ b/cmd/doctor.go
@@ -225,11 +225,11 @@ func doctorFindings(fleetHome string) ([]doctorFinding, error) {
 	}
 	for _, task := range tasks {
 		briefPath := filepath.Join(fleetHome, task.Brief)
-		data, err := os.ReadFile(briefPath)
-		if err != nil && !os.IsNotExist(err) {
+		declares, err := briefDeclaresReportPath(briefPath, state.ReportPath(fleetHome, task.ID))
+		if err != nil {
 			return nil, fmt.Errorf("read task %q brief: %w", task.ID, err)
 		}
-		if !strings.Contains(string(data), state.ReportPath(fleetHome, task.ID)) {
+		if !declares {
 			findings = append(findings, doctorFinding{Severity: doctorWarning, Text: fmt.Sprintf("task %q brief does not declare report path", task.ID)})
 		}
 	}
@@ -277,6 +277,42 @@ func doctorFindings(fleetHome string) ([]doctorFinding, error) {
 		}
 	}
 	return findings, nil
+}
+
+func briefDeclaresReportPath(briefPath, reportPath string) (bool, error) {
+	data, err := os.ReadFile(briefPath)
+	if err != nil {
+		info, statErr := os.Stat(briefPath)
+		if os.IsNotExist(err) || (statErr == nil && info.IsDir()) {
+			return false, nil
+		}
+		return false, err
+	}
+	text := string(data)
+	for start := 0; start < len(text); {
+		at := strings.Index(text[start:], reportPath)
+		if at < 0 {
+			return false, nil
+		}
+		end := start + at + len(reportPath)
+		if end == len(text) || !reportPathSuffix(text, end) {
+			return true, nil
+		}
+		start += at + 1
+	}
+	return false, nil
+}
+
+func reportPathSuffix(text string, end int) bool {
+	next := text[end]
+	if next == '.' && end+1 < len(text) && isReportPathSuffixChar(text[end+1]) {
+		return true
+	}
+	return isReportPathSuffixChar(next)
+}
+
+func isReportPathSuffixChar(char byte) bool {
+	return (char >= 'a' && char <= 'z') || (char >= 'A' && char <= 'Z') || (char >= '0' && char <= '9') || char == '_' || char == '-' || char == '/' || char == '\\'
 }
 
 // A local-only fleet never needs gh, while a registered project delivering through direct-pr or
--- a/cmd/doctor.go
+++ b/cmd/doctor.go
@@ -90,7 +90,11 @@ func newDoctorCmd(info selfupdate.BuildInfo) *cobra.Command {
 			if err != nil {
 				return asPrecondition(err)
 			}
-			findings, err := doctorFindings(fleetHome)
+			histories, err := state.ListOpenHistoriesReadOnly(fleetHome)
+			if err != nil {
+				return err
+			}
+			findings, err := doctorFindings(fleetHome, histories)
 			if err != nil {
 				return err
 			}
@@ -185,7 +189,7 @@ func legacyDoctorCompatibility() bool {
 	return legacyDoctorCompat
 }
 
-func doctorFindings(fleetHome string) ([]doctorFinding, error) {
+func doctorFindings(fleetHome string, histories []state.TaskHistory) ([]doctorFinding, error) {
 	findings := make([]doctorFinding, 0)
 
 	agentsViolations, err := agentsmd.Check(fleetHome)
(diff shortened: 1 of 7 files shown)


atqamz:
Supervisor review of PR #409. The shape is right, the before/after prompt is exactly what I asked for, and both `buildHarness` call sites are covered - I checked, there are only two. One thing to fix before this can merge.

`briefPrompt` appends the report sentence unconditionally, so an empty `Options.ReportPath` produces the string "The worker report channel is ." and then instructs the worker to append to nothing. That is the exact failure class this issue exists to remove: a silently wrong instruction that looks like a correct one. Production cannot reach it today, but `harness.Options` is an exported struct in an exported package and nothing stops a future caller from omitting the field, which is how the original defect happened - the contract was written down and nothing enforced it.

Make it structural, not documented. Either have `briefPrompt` refuse to build a prompt without a report path (an error the caller must handle, matching how you already handle the `filepath.Abs` failure in `absoluteReportPath`), or drop the sentence entirely when the path is empty so no false instruction is ever issued. I prefer the first: an unreportable launch is not a launch worth making. Add a test for whichever you choose.

While you are in there, confirm one thing in the PR body: does the injected path survive `hand reopen` as well as `hand spawn`? You patched `reconcile.go:1339` for launch confirmation, and reopen goes through `provision` so it should be covered, but say so explicitly - reopen is how a mute task gets rescued and it is the path I will actually use on the nine already-broken tasks.

Do not merge and do not close the issue. Merge and close are the supervisor's action only. Report progress to /home/atqa/github/atqamz/hand/state/406-brief-report-channel.status as you have been.

atqamz:
Reviewed PR atqamz/hand#409 at `1a8ec39c26`. The empty-`ReportPath` guard is what I asked for and better than what I asked for: pushing the error up through `Build`'s switch means every prompt-capable builder fails closed rather than each one checking, and `TestBuildRejectsMissingReportPath` pins it. `absoluteReportPath` in both `provision.go` and `reconcile.go` covers spawn and reopen. Four things before this merges - two of them P1, and the first is not visible from inside your own PR.

**P1 - you and PR atqamz/hand#404 are both adding a state-database open to `doctorFindings`, and whichever merges second inherits the other's cost.**

I measured all three on the same `windows-latest` runner:

| branch | `cmd` | `internal/runtime` |
| --- | --- | --- |
| `main` (run 32952412368) | 353.6s | 217.0s |
| this PR (run 32955808867) | **380.8s** | 256.4s |
| atqamz/hand#404 (run 32955191776) | **timeout at 600s** | 219.8s |

Go's default per-package test timeout is 600s, so `main` already spends 59% of it in `cmd` and leaves
246s of headroom. Your `state.ListOpen(fleetHome)` call plus three doctor tests take 27s of that.
atqamz/hand#404 adds `state.ListOpenHistoriesReadOnly(fleetHome)` to the same function plus three more
doctor tests, and takes the rest. Neither of you is doing anything unreasonable; you are independently
spending one shared budget, and each PR passes alone.

Two consequences.

First, coordinate the open. `doctorFindings` must open the store **once** and pass what it read to both
checks - yours and atqamz/hand#404's worktree check - not open it twice. Whichever of you merges first
should leave the seam for the other. Say in your PR body which shape you are leaving.

Second, use a read-only path. `state.ListOpen` runs the full open, which triggers registry migration
and its journal fsync; on Windows that fsync is the entire cost. `project.ListReadOnly` sits eight
lines below yours in the same function and was deliberately chosen for exactly this reason. There is no
`ListOpenReadOnly` for `[]Task` today - only `ListOpenHistoriesReadOnly` - so either add one or read
the histories atqamz/hand#404 is already reading and take `.Task` from them. `hand doctor` is a
diagnostic; it must not migrate anything.

The 600s cap itself I filed separately as atqamz/hand#416. Do not raise it here.

**P1 - one unreadable brief bricks `hand doctor` entirely.**

```go
declares, err := briefDeclaresReportPath(briefPath, state.ReportPath(fleetHome, task.ID))
if err != nil {
    return nil, fmt.Errorf("read task %q brief: %w", task.ID, err)
}
```

`briefDeclaresReportPath` swallows not-exist and is-a-directory, and returns every other error - a
permission bit, a bad symlink, an I/O error on one brief among twenty. `doctorFindings` then aborts and
`hand doctor` produces nothing at all. This is the command an operator reaches for *because* something
is wrong with the home, so it has to be the most degradation-tolerant thing in the tree. Turn an
unreadable brief into a finding that names the task and the read error, and keep going.

**P2 - the report path is compared unnormalized, so the check can warn on a correct brief.**

`state.ReportPath` is `filepath.Join(Dir(homeDir), id+".status")` with no `Abs`. Your own
`absoluteReportPath` wraps it in `filepath.Abs` precisely because `fleetHome` is not guaranteed
absolute - and then the doctor check calls the raw form. If `fleetHome` arrives relative, you compare a
relative path against briefs that (correctly) carry the absolute one, `strings.Index` misses, and
`hand doctor` warns "task %q brief does not declare report path" for every open task while every brief
is fine. Use the same normalization on both sides.

While you are in `briefDeclaresReportPath`: the suffix logic reads correctly to me - a trailing `.`
followed by an alnum rejects `.status.bak`, a bare trailing `.` or newline accepts - but it is doing
by hand what a word-boundary check does, and it is the kind of code that is right today and wrong after
the next edit. A short comment stating the rule it implements would earn its place here.

**P2 - required in the PR body: reference atqamz/hand#418, and scope your claim honestly.**

This PR does not fix atqamz/hand#406 for `grok` or `pi`. Both builders pass the brief as a file and no
prompt at all, so `briefPrompt` is unreachable from either and a worker on those harnesses still has no
report channel. `promptCapable` lists four of six.

That is not a defect you introduced and I am not asking you to fix it here. But `Fixes atqamz/hand#406`
on this PR will read, in the merged history, as "the report channel reaches workers", and for two
harnesses it does not. State the limitation in the PR body and point at atqamz/hand#418, which I filed
off this review.

Related: while you are there, `CarriesPrompt`'s doc comment cites atqamz/hand#36 for the missing prompt
content. atqamz/hand#36 is closed and was about herdr agent detection, not prompt carriage. Repoint it
at atqamz/hand#418 - that is a one-line fix and it belongs with this change, because you are the reason
the consequence got worse: it used to mean a grok worker missed the operator-decision rule, and now it
means a grok worker cannot speak at all.

**Nit - the doc comment above `briefPrompt` is detached.**

You left a blank line between the three-line comment block and `func briefPrompt`, so it is no longer
that function's doc comment. Close the gap.

Do not merge and do not close the issue: merge and close are the supervisor's action only.

PR #414, opened 2026-08-26
respect task holds during orientation

### LINKED ISSUE #408: hand orient ignores holds, so a deliberately held task pins next_action forever
## What happens

`hand hold set` records what a task is waiting on and why, and `hand status` shows it. `hand orient`
ignores it completely: the task keeps emitting `parked` and `unreported` actionable subjects, and
because `next_action` picks the oldest actionable item, orientation locks onto a held task and names
the same one every turn, forever.

## Reproduction, this fleet home

Three `omanixy` tasks (`23-launcher-activation`, `24-font-wallpaper-ownership`, `39-tray-sni-proof`)
have been `state: idle` for five hours with `flags: unreported parked`. Both flags are explained: the
briefs carry no report channel (atqamz/hand#406), so no report was ever possible, and their worktrees
are an instance of atqamz/hand#403. I recorded exactly that with a hold:

```
$ hand hold set 23-launcher-activation --kind blocked \
    --blocked-on 403-cross-fleet-worktree-resolution --reason "..."
blocked_on: 403-cross-fleet-worktree-resolution

$ hand status 23-launcher-activation
state: idle
held: "waiting on 403-cross-fleet-worktree-resolution: Mute by construction (atqamz/hand#406) ..."
flags: unreported parked
```

The hold is stored and displayed. Then:

```
$ hand orient
next_action_kind: unreported
next_action_task: 23-launcher-activation
next_action_reason: Run `hand status 23-launcher-activation`; it stopped without reporting a terminal state
```

Identical before and after the hold. Same for all three tasks, and the actionable list still carries
`parked` + `unreported` for each.

## Where it comes from

`internal/attention/attention.go` has no notion of a hold. `Evidence` carries `Unannounced`,
`Unreported`, `Parked`, `Repair`, `SendUncertain`, `GateProblem` and the rest - there is no `Held`
field, so `Derive` cannot consider one. `cmd/session.go:174` does load holds alongside views, and
passes them to `classifyNextAction` at line 191 and `appendFleetState` at line 216, so the data is
right there at the call site; the attention layer that decides `Actionable` never sees it.

## Why this matters more than it looks

`AGENTS.md` makes `hand orient` mandatory at the start of every supervisor turn, and the whole point
of `next_action` is to be the queue. A held task pins that queue to work the supervisor has already
adjudicated and deliberately deferred, which trains the supervisor to ignore `next_action` - and the
one turn it names something new is the turn that gets ignored too. It also means a supervisor's own
recorded decision is invisible to the tool that is supposed to summarise fleet state, so the same
reasoning gets redone every turn.

Note that `unreported` here is not even a real signal: per atqamz/hand#406 these workers had no channel
to report on. So orientation is insisting on an action for a condition that is both explained and
impossible to clear.

## Fix direction

1. Add a `Held` (or `HoldKind`/`HoldReason`) field to `attention.Evidence` and populate it from the
   holds `cmd/session.go` already loads.
2. A hold should not erase a subject - the condition is still true and still worth reporting - it
   should make it **not actionable**. Keep emitting `parked`/`unreported` in the actionable list's
   provenance if that is useful, but stop them from being eligible for `next_action`, the same way
   `KindUnannounced` is already emitted with `actionable: false`.
3. Decide deliberately how `--kind operator` differs from `--kind blocked` here. An operator hold is
   arguably *waiting on a human* and should be visible-but-not-actionable; a `blocked` hold names
   another task and can clear itself when that task lands. Say which in the implementation rather
   than treating both the same by accident.
4. When every subject on a task is held, `hand orient` should say so rather than falling silent about
   the task - something the supervisor can read as "deferred by you, still true".

## Acceptance

- A held task does not become `next_action`.
- Clearing the hold makes it eligible again with no other change.
- `hand orient` still reports the underlying condition, so a hold hides nothing.
- An `operator`-kind hold and a `blocked`-kind hold each behave as the implementation explicitly
  chose, with a test for each.

Related: atqamz/hand#406

atqamz:
## Summary
- Propagate task holds into shared attention evidence and exclude held subjects from next-action ranking.
- Preserve held conditions in a dedicated orientation_deferred surface, including operator and blocked hold context.

Fixes atqamz/hand#408

## Test plan
- `make test`
- `make lint`
- Cross-build with CGO disabled for linux amd64/arm64, darwin amd64/arm64, and windows amd64.

## Summary
- Propagate task holds into shared attention evidence and exclude held subjects from every next-action branch.
- Preserve held conditions in bounded `orientation_deferred` rows with the subject kind, hold kind, and `blocked_on` facts; full free-text reasons remain in `hand status`.
- Held subjects are silent by design once deferred. Resolution of a satisfied `blocked_on` is tracked in atqamz/hand#417; this PR deliberately does not close that loop.

Fixes atqamz/hand#408

## Test plan
- `make test`
- `make lint`
- Cross-build with CGO disabled for linux amd64/arm64, darwin amd64/arm64, and windows amd64.

### DIFF
--- a/cmd/evidence.go
+++ b/cmd/evidence.go
@@ -128,7 +128,7 @@ func appendTaskAttention(evidence *orientation.Evidence, view taskView, targetEv
 		} else if view.held {
 			evidence.Deferred = append(evidence.Deferred, orientation.DeferredSubject{
 				TargetID: view.task.ID,
-				Kind:     subject.Kind, Reason: fmt.Sprintf("%s; deferred by %s hold: %s", item.Reason, view.hold.Kind, holdDetail(view.hold)),
+				Kind:     subject.Kind, HoldKind: view.hold.Kind, BlockedOn: view.hold.BlockedOn,
 				Provenance: subject.Provenance,
 			})
 		}
--- a/cmd/evidence.go
+++ b/cmd/evidence.go
@@ -94,15 +94,7 @@ func (s *fleetSnapshot) evidence() orientation.Evidence {
 			evidence.Targets = append(evidence.Targets, targetEvidence)
 		}
 		evidence.Work = append(evidence.Work, orientation.WorkEvidence{ID: view.task.ID, Kind: "task", State: view.agentState, Reported: view.reportedState})
-		for _, subject := range attention.Derive(taskAttentionEvidence(view)) {
-			if !subject.Actionable {
-				continue
-			}
-			evidence.Actionable = append(evidence.Actionable, orientation.ActionableEvidence{
-				TargetID: view.task.ID, TargetKind: "task", Generation: targetEvidence.Generation,
-				Kind: subject.Kind, Reason: subject.Reason, Provenance: subject.Provenance,
-			})
-		}
+		appendTaskAttention(&evidence, view, targetEvidence)
 	}
 	if s.next.Kind != "" {
 		evidence.NextActions = append(evidence.NextActions, orientation.NextAction{Kind: s.next.Kind, Target: s.next.Task, Command: s.next.Command, Reason: s.next.Reason})
@@ -123,4 +115,22 @@ func (s *fleetSnapshot) evidence() orientation.Evidence {
 		})
 	}
 	return evidence
+}
+
+func appendTaskAttention(evidence *orientation.Evidence, view taskView, targetEvidence orientation.TargetEvidence) {
+	for _, subject := range attention.Derive(taskAttentionEvidence(view)) {
+		item := orientation.ActionableEvidence{
+			TargetID: view.task.ID, TargetKind: "task", Generation: targetEvidence.Generation,
+			Kind: subject.Kind, Reason: subject.Reason, Provenance: subject.Provenance,
+		}
+		if subject.Actionable {
+			evidence.Actionable = append(evidence.Actionable, item)
+		} else if view.held {
+			evidence.Deferred = append(evidence.Deferred, orientation.DeferredSubject{
+				TargetID: view.task.ID,
+				Kind:     subject.Kind, Reason: fmt.Sprintf("%s; deferred by %s hold: %s", item.Reason, view.hold.Kind, holdDetail(view.hold)),
+				Provenance: subject.Provenance,
+			})
+		}
+	}
 }
--- a/cmd/nextaction.go
+++ b/cmd/nextaction.go
@@ -96,13 +96,7 @@ func classifyNextAction(cfg workerConfig, projectCount int, backlog backlogSumma
 		return nextAction{Kind: nextActionUnacknowledged, Task: v.task.ID, Command: statusCommand(v.task.ID),
 			Reason: statusReason(v.task.ID, "act on its unacknowledged worker event")}
 	}
-	if h, ok := firstHold(holds); ok {
-		for _, v := range views {
-			if v.task.ID == h.ID {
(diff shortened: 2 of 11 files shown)


atqamz:
Reviewed PR atqamz/hand#414 at `af693f583d`. The shape is right: `Held` on `attention.Evidence`, blanket suppression in `Derive`, `hasAttentionKind` requiring `Actionable` so all fourteen ranking branches skip held tasks, and `firstOrphanHold` keeping the orphan case that `firstHold` used to conflate. Deferring instead of dropping is the correct call. One blocker, one required PR-body change, one nit.

**P1 - `orientation_deferred` is the only unbounded field in a surface whose entire contract is boundedness.**

`bound()` in `internal/orientation/orientation.go` applies `boundedText` (256 runes) field-by-field to `Work`, `Actionable`, `NextActions` and `Errors`. Your change adds the `maxOrientationItems` count truncation for `Deferred` but no `for i := range result.Deferred` text pass. And `Deferred.Reason` is the longest string on the surface by a wide margin, because `appendTaskAttention` builds it as:

    fmt.Sprintf("%s; deferred by %s hold: %s", item.Reason, view.hold.Kind, holdDetail(view.hold))

`holdDetail` returns the operator's free text verbatim - `fmt.Sprintf("waiting on %s: %s", h.BlockedOn, h.Reason)` for a blocked hold, plus a `" (inferred from a pane scrape)"` suffix. `state.SetHold` validates the id and `blocked_on` and never bounds `Reason`.

Measured against this fleet home's live holds, not hypothetically. My hold detail on `401-mechanical-staleness-fix` is ~1,200 characters; the three omanixy holds are ~700 each. 401 currently carries two non-actionable subjects (`parked`, `unreported`), so under this PR it emits two rows of ~1,250 characters, the omanixy tasks add three more, and `hand orient` grows by roughly 5KB while `orientation_truncated` still reports `false`. That is the exact failure the bound exists to prevent, and a supervisor reading orient would be told the surface was complete when it had silently blown its budget.

Do not fix this by adding `Deferred` to the `boundedText` loop and stopping there. A deferred row's job is "this subject exists and a hold is deferring it", not "reproduce the hold". Put the structured, short facts in the row - hold kind, and `blocked_on` when there is one - and leave the free-text reason to `hand status`, which is the unbounded surface and already prints it. Then add the `boundedText` pass anyway, so the field is bounded by construction rather than by the current formatting happening to be short.

**Required in the PR body: reference atqamz/hand#417.**

I filed it off this review. Your change is a strict improvement, and it also converts a false positive into a false negative: before it, a held task surfaced in `orientation_actionable` mislabelled `parked`; after it, held tasks are excluded from every next-action branch and appear only in `orientation_deferred`. Nothing in Hand ever reads `blocked_on` as a dependency - it is validated, stored, and rendered as text, and no code compares it to the blocker's state - so a `blocked` hold is now write-only. Four tasks in this home are held on `403-cross-fleet-worktree-resolution` and nothing will tell me when it lands.

That is out of scope for atqamz/hand#408 and I am not asking you to build it here. State it in the PR body: that held subjects are now silent by design, that resolution of a satisfied `blocked_on` is tracked in atqamz/hand#417, and that this PR deliberately does not close that loop. A reader of the merged history has to be able to see the gap was known.

**Nit - `sortedDeferred` uses `sort.Slice`; every sibling uses `sort.SliceStable`.**

The (TargetID, Kind, Reason) tiebreak chain makes it deterministic unless two rows match on all three, and then the order is unspecified. Two subjects on one task can collide once you shorten `Reason` per the P1 above. Match the siblings.

Do not merge and do not close the issue: merge and close are the supervisor's action only.

PR #415, opened 2026-08-26
Reconcile landing evidence and terminal cleanup

### LINKED ISSUE #407: reconcile asserts unlanded for a ship task with no recorded PR, so merged work converges to interrupted
## What happens

`hand reconcile` treats "this task has no recorded PR" as **positive evidence that the work did not
land**, and converges the attempt to `interrupted`. `internal/runtime/reconcile.go:509`:

```go
if task.PR == "" {
    return landingUnlanded
}
```

Everywhere else in the same function, absence of evidence yields `landingUnknown` - a merge state
that cannot be read returns `unknown`, a scout report that cannot be stat'ed returns `unknown`. Only
the empty-PR case invents a negative answer.

`hand reconcile --help` states what `unlanded` means: "positively observed unlanded work interrupts
it". Nothing was positively observed here. Nothing was observed at all.

## Live reproduction, this fleet home

Task `402-report-pipeline-fix`, a ship task whose PR atqamz/hand#405 merged at 2026-08-26T09:19:45Z
and whose branch was deleted on merge:

```
$ hand reconcile 402-report-pipeline-fix
result: healthy
action: converge-terminal-lifecycle
landing: unlanded
liveness: idle-unreported
detail: attempt 78 converged to interrupted on observed evidence

$ gh pr list --repo atqamz/hand --head 402-report-pipeline-fix --state all \
    --json number,state,mergedAt
#405 MERGED 2026-08-26T09:19:45Z
```

The work is merged into `main`. hand recorded the attempt as `interrupted` "on observed evidence",
and the worktree (`/home/atqa/.treehouse/hand-a7a4d5/12/hand`) is still held.

## Why hand could have known

hand already owns branch-based PR discovery that finds merged PRs specifically:
`ghutil.ObservePRByBranch` (`internal/ghutil/pr.go:98`), whose query is built at
`internal/ghutil/pr.go:162` with `--state all`, and whose own comment at line 158 says `--state all`
is there "because gh pr list defaults to open only and a gate-opened PR may already be" merged.
`internal/faketool/FIDELITY.md:292` says it even more directly: "`--state all` is what makes a merged
PR findable at all."

`observeLanding` never calls it. It reads `task.PR` and gives up.

Worth noting how the task got into this state, because it is not an unusual path: the PR was opened
by the worker, not by the supervisor, so nothing ran `hand pr <id> <url>`. `hand status` did display
it for a while as `pr: none (observed only: <url>)` - hand saw the PR out of band and printed it -
and once the branch was deleted on merge even that annotation disappeared. So the supervisor is
expected by convention to run `hand pr` and nothing enforces or reminds, which is structurally the
same failure as atqamz/hand#406.

## Fix direction

1. `observeLanding` should return `landingUnknown`, not `landingUnlanded`, when a ship task has no
   recorded PR. That alone stops a landed task being recorded as interrupted "on observed evidence"
   and lets the existing needs-repair path say so honestly.
2. Better: before giving up, try `ghutil.ObservePRByBranch` on the attempt's branch. hand knows the
   task's branch and its project's repo, the query already exists, and it is the one query that
   finds a merged-and-deleted branch. A found merged PR is real landing evidence.
3. Consider persisting what the "observed only" path already discovers, so a PR hand has seen once
   does not become invisible when the branch is deleted.

## Not in scope

Do not change the second `task.PR == ""` branch at `internal/runtime/reconcile.go:1567`. That one is
correct: it *withholds* the worktree return rather than asserting anything, which is the safe
direction.

## Acceptance

- A ship task with no recorded PR whose branch has a merged PR reconciles to a landed, completed
  attempt.
- A ship task with no recorded PR and no discoverable PR reconciles to `unknown`, not `unlanded`.
- `landing: unlanded` is reachable only from an actual observation.

Related: atqamz/hand#406

atqamz:
## Summary
- discover and persist a ship PR by attempt branch when no PR is recorded
- return unknown when landing cannot be observed, avoiding false unlanded conclusions
- let teardown release resources held by terminal tasks only after commit safety is proven

## Verification
- `make lint`
- `make build`
- focused runtime tests and race tests
- serial runtime suite
- cross-platform compile/build checks for linux/amd64, linux/arm64, darwin/arm64, darwin/amd64, and windows/amd64
- `make test` was attempted with a 180-second limit and hit the existing toolchain extraction timeout in `TestSpawnLegacyTierCreatesAttemptBeforeWaitingForProjectLock` before completion

Closes #407

## Summary
- distinguish proven PR absence from unknown observation during reconciliation
- persist discovered PRs and propagate project or persistence failures
- safely clean resources held by terminal tasks, including provisioning attempts, with pre-release lifecycle and commit-safety validation

## Verification
- `make test` passes with race detection across all packages
- `make lint`
- `make build`
- focused reviewer regression tests pass under race detection
- cross-platform compile/build checks pass for linux/amd64, linux/arm64, darwin/arm64, darwin/amd64, and windows/amd64

Closes #407

### DIFF
--- a/internal/runtime/pr.go
+++ b/internal/runtime/pr.go
@@ -43,9 +43,13 @@ func observePR(ctx context.Context, homeDir, recordedPR string, active state.Att
 	// Live worktree state wins when it is determinable, since it is more current than what attempt
 	// creation recorded; the durably stored branch is the fallback, since a detached or torn-down
 	// worktree cannot answer at all - not evidence that no branch, and so no PR, ever existed.
-	branch, liveErr := currentBranch(active.Worktree)
-	if liveErr != nil {
-		branch = active.Branch
+	branch := active.Branch
+	var liveErr error
+	if active.Worktree != "" {
+		branch, liveErr = currentBranch(active.Worktree)
+		if liveErr != nil {
+			branch = active.Branch
+		}
 	}
 	if branch == "" {
 		return ghutil.UnknownPRObservation("git symbolic-ref --short -q HEAD", fmt.Sprintf("resolve the branch to search for in %s: %v", active.Worktree, liveErr))
--- a/internal/runtime/reconcile.go
+++ b/internal/runtime/reconcile.go
@@ -513,11 +513,26 @@ func (r *Runtime) observeLanding(ctx context.Context, home string, task state.Ta
 	}
 	if task.PR == "" {
 		projectInfo, found, err := project.FindReadOnly(home, task.Project)
-		if err != nil || !found {
-			return task, landingUnknown, nil
+		if err != nil {
+			return task, landingUnknown, fmt.Errorf("observe landing project %q: %w", task.Project, err)
+		}
+		if !found {
+			if task.Kind == "" {
+				return task, landingUnlanded, nil
+			}
+			return task, landingUnknown, fmt.Errorf("observe landing project %q: project is not registered", task.Project)
+		}
+		if projectInfo.Mode == project.ModeLocalOnly && (task.Kind == "" || (attempt.Worktree != "" && attempt.Branch == "")) {
+			return task, landingUnlanded, nil
 		}
 		detected, observation, err := DetectPR(ctx, home, task, attempt, projectInfo)
-		if err != nil || !observation.Found() {
+		if err != nil {
+			return task, landingUnknown, fmt.Errorf("record observed PR for task %q: %w", task.ID, err)
+		}
+		if observation.Absent() {
+			return task, landingUnlanded, nil
+		}
+		if observation.Unknown() {
 			return task, landingUnknown, nil
 		}
 		if observation.Merged && !task.MergeAnnounced {
--- a/internal/runtime/reconcile.go
+++ b/internal/runtime/reconcile.go
@@ -350,7 +350,12 @@ func (r *Runtime) reconcileTask(ctx context.Context, home, id string, attest rec
 			return result, err
 		}
 		if terminalConvergenceCandidate(attempt, observation) {
-			observation.Landing = r.observeLanding(ctx, home, history.Task, mergeObserved)
+			history.Task, observation.Landing, err = r.observeLanding(ctx, home, history.Task, attempt, mergeObserved)
+			if err != nil {
+				result.Outcome = reconcileOutcomeBlocked
+				result.Landing = string(observation.Landing)
+				return result, err
+			}
 			result.Landing = string(observation.Landing)
 		}
 		decision := decideReconciliation(history.Task, attempt, observation)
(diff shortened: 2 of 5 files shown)


atqamz:
Reviewed PR atqamz/hand#415 at `be8f96421a`. You reported `done:` with "CI checks are queued"; they have since come back and five Test jobs plus the Nix build are red. Read this before you touch them, because the failures are telling you the right answer and the tempting fix is the wrong one.

**P1 - the three failing tests are correct and your change is wrong. Do not change the tests.**

```
--- FAIL: TestReconcileConvergesDonePaneWithoutWorkerReport
    liveness_test.go:86: result = {... Landing:unknown ...}, want unlanded landing evidence
--- FAIL: TestReconcileCommandRendersConvergedTerminalLifecycle
    reconcile_test.go:85: task "task-1" needs repair: running-pane-missing
--- FAIL: TestReconcileRunningMissingPaneConvergesWithoutReplacement
```

All three, on all five platforms, are one cause:

```go
detected, observation, err := DetectPR(ctx, home, task, attempt, projectInfo)
if err != nil || !observation.Found() {
    return task, landingUnknown, nil
}
```

`!observation.Found()` is true for **two** different observations, and `ghutil` deliberately keeps them apart. `ObservePRByBranch` returns `absentPR(probe)` when `len(results) == 0` - every target answered and none carries that head ref - and `Absent`'s own doc says so in as many words:

> Absent reports that the query completed and there is genuinely no such pull request. It is never the residue of a failed query, so a caller may act on it.

A proven absence is not an unobservable state. A ship task whose branch genuinely has no PR **is** unlanded, and that is the one case the code you replaced got right. By collapsing `Absent` into `landingUnknown` you fixed atqamz/hand#407's false `unlanded` by installing its mirror image: a false `unknown` that withholds resource release forever on a task nothing will ever come back for. The failing fixtures are exactly that shape - no PR, absence proven - and they assert `unlanded`.

The four-way mapping you want:

| observation | landing |
| --- | --- |
| `Found()` and `Merged` | `landingLanded` |
| `Found()` and not merged | `landingUnlanded` |
| `Absent()` | `landingUnlanded` |
| `Unknown()` | `landingUnknown` |

Use `observation.Absent()` and `observation.Unknown()` explicitly rather than negating `Found()`. Ambiguous results already come back as unknown, which is right - two candidate PRs is not evidence of anything.

Do the same audit on the two error swallows in that block. `project.FindReadOnly` failing and the project not being registered are different facts, and `DetectPR`'s error is a *persistence* failure from `RecordPR` after a successful observation - hand discovered a PR and failed to write it down. Returning bare `landingUnknown, nil` for all of these loses the reason. At minimum, do not conflate "could not ask" with "asked and the answer was no".

**P1 - `finishTeardown` can now fail after the worktree and Herdr pane are already gone.**

```go
} else if active.Lifecycle != terminalAttempt {
    return Result{}, WithWarnings(fmt.Errorf("terminal task %q attempt %d has lifecycle %q, want recorded teardown lifecycle %q", ...), warnings)
}
```

`finishTeardown` runs after `releaseHerdr` and after `releaseWorktree`. So on this path the attempt keeps a non-terminal teardown lifecycle while both its resources are irreversibly released, and every later `hand teardown` re-enters `terminalAttemptWithResources`, finds nothing unsettled, and reports "no active attempt". The task is then unreachable by the very command meant to clean it up - which is the failure atqamz/hand#407 exists to remove, reintroduced one layer down.

A precondition that can only fail after irreversible work has to run before the irreversible work. Either validate the attempt lifecycle up where you compute `terminalCleanup`, or make this a warning and terminalize anyway. It must not be a hard error at this point in the sequence.

**P2 - `terminalAttemptWithResources` skips provisioning and running attempts, and a provisioning attempt can hold a lease.**

```go
if attempt.Lifecycle == state.AttemptProvisioning || attempt.Lifecycle == state.AttemptRunning {
    continue
}
```

For `running` that is clearly right. For `provisioning` I am not convinced: an attempt that acquired a worktree lease and then died before launching is precisely a stranded-resource shape, and this skips it. Either include it, or say in a comment why a provisioning attempt's lease is somebody else's problem. Do not leave the reader to guess.

**Good, and worth keeping exactly as it is:** the commit-safety gate before releasing a terminal task's worktree. `commitSafetyProvenDurable` or refuse, with the reason in the message, is the right shape - it is the difference between reclaiming a slot and losing a worker's unlanded work. The `recoveredCompletion` restructure also reads correctly to me: the early return sits after both releases, so a recovered record still finishes cleanly.

**Notes, not blockers.**

Your branch is `fix/407-reconcile-landing-evidence`. The repository convention is `<issue#>-<slug>`, so `407-reconcile-landing-evidence`. Not worth a force-push mid-review; match it next time.

Windows `cmd` came in at 600.652s on your run - the same 10-minute cap that has now bitten three PRs today. That is not yours to fix; it is filed as atqamz/hand#416. Do not raise the timeout and do not treat a Windows `cmd` timeout as evidence about your change until the three real failures above are gone.

On `make test`: your PR body says it hit the `TestSpawnLegacyTierCreatesAttemptBeforeWaitingForProjectLock` toolchain-extraction timeout at 180s and never completed. That is honest reporting and I would rather have it than a false green - but it means `done:` rested on focused tests while three suite-level tests were failing. CI is the gate here, not the focused subset.

Do not merge and do not close the issue: merge and close are the supervisor's action only. `done:` never means merged, and it does not mean green either - name the run you are calling green when you report.

PR #434, opened 2026-08-27
test(store): reduce fsync cost for test databases

### LINKED ISSUE #416: Windows CI has no timeout headroom: cmd uses 59% of the 10m cap on main, so the next slow PR panics with a misleading dump
## What happens

On `main`, the `cmd` package consumes 59% of Go's default 10-minute per-package test timeout on
`windows-latest`. Measured on the last green `main` run (32952412368, commit on `main`):

| platform | `cmd` | `internal/runtime` | `internal/watcher` |
| --- | --- | --- | --- |
| ubuntu-latest | 144.1s | 102.3s | - |
| macos-15-intel | 182.3s | 112.9s | - |
| **windows-latest** | **353.6s** | **217.0s** | 289.1s |

That leaves 246 seconds of headroom in `cmd` before `panic: test timed out after 10m0s`. Windows is
~2.5x slower per test than Linux on the same suite, and the cost is fsync-bound rather than CPU-bound:
every state-database write pays a real `FlushFileBuffers`, which on a GitHub Windows runner is
expensive.

## Why it matters

A PR that adds a handful of database-touching tests to `cmd` does not fail with "your new tests are
slow". It fails with a `panic: test timed out after 10m0s` whose goroutine dump names whichever test
happened to be running when the alarm fired - a test that had been running for one second and is not
hung at all. The dump shows a single goroutine blocked in a `modernc.org/sqlite` fsync under
`database/sql`, which reads exactly like a deadlock or an unclosed connection.

This happened on atqamz/hand#404. Its worker and I both spent real time chasing a nonexistent lock
bug on the strength of that dump. The actual cause was three new `doctor` tests plus an added
state-database open on every `doctor` invocation, against a package that was already at 353s.

`internal/watcher` at 289s and `internal/runtime` at 217s are on the same trajectory.

## Fix direction

Pick one; the first is the smallest.

1. **Set an explicit, generous `-timeout` in the Makefile's test target** so the failure mode for a
   slow package is a slow job rather than a misleading panic. This does not fix the slowness, but it
   stops the slowness from being misdiagnosed as a hang. Cheap and immediate.
2. **Cut the per-test fsync cost.** The `cmd` suite opens and migrates a state database many times
   over. Options: open it lazily so a command that has nothing to read never opens it, share one
   migrated database across tests that do not mutate it, or set `PRAGMA synchronous=OFF` in the test
   harness only (never in production) so journal writes do not fsync.
3. **Report the budget in CI.** Have the Test job fail, or at least annotate, when any package exceeds
   a fraction of its timeout - say 70%. That turns "we are 40 seconds from a mystery panic" into a
   visible warning on the PR that consumed the margin, not on the one after it.

I would take 1 now and 2 as the real fix.

## Acceptance

- A package that runs long fails with a clear duration signal rather than a `test timed out` panic
  naming an unrelated test.
- Windows `cmd` and `internal/watcher` each sit below 60% of their timeout on `main`, with the
  measurement recorded so the next regression is visible.

Related: atqamz/hand#404

### LINKED ISSUE #430: test: set an explicit -timeout so a slow package fails with its duration instead of a misleading panic
## What happens

`make test` and CI's `Test` job both run `go test -tags=test -race ./...` with no `-timeout`, so every
package inherits Go's 10-minute default. When a package crosses it the failure is
`panic: test timed out after 10m0s` plus a goroutine dump naming whichever test the alarm happened to
interrupt.

That has already happened on `main`. Run 32963174452 (2026-08-26T11:24, `windows-latest`):

```
panic: test timed out after 10m0s
	running tests:
		TestSpawnRefusedAtTheTrustPromptUnwindsAndSpawnsAgain (2s)
...
FAIL	github.com/atqamz/hand/cmd	600.214s
FAIL	github.com/atqamz/hand/internal/runtime	600.583s
```

The named test had been running for two seconds and was not hung. The only informative goroutine is
blocked in a `modernc.org/sqlite` fsync under `database/sql`, which reads like a lock bug and is not
one.

## Why this is worth its own change

atqamz/hand#416 owns the real repair - cutting the fsync cost - and is on 0.8.0. This is the
narrow half that stops the *diagnosis* from being wrong, costs almost nothing, and is worth having
before the 0.7.0 release, because a red `main` from a mystery panic is expensive twice: once in CI
time and once in whoever chases the phantom deadlock. It has already cost that on atqamz/hand#404
and again on atqamz/hand#401.

Today's green `main` (run 33049302071) leaves `cmd` at 431.3s of the 600s cap on `windows-latest`, up
from 353.6s when #416 was filed. 168 seconds of margin, trending the wrong way, so this will fire
again.

## Scope

- Set an explicit `-timeout` on the unit-test invocation in the `Makefile` and in
  `.github/workflows/ci.yaml`, generous enough that crossing it means something is genuinely wrong
  rather than merely slow. The e2e and contract targets already pass `-timeout=10m` explicitly; this
  is about the `./...` unit run.
- Do **not** change any pragma, test, or database behaviour here. Making the suite faster is #416.
- Do not silence or skip the slow packages.

## Acceptance

- A package that takes longer than Go's default 10 minutes reports a slow package with its duration,
  not `panic: test timed out`.
- The value chosen is written down with its reasoning, so the next person raising it knows what it was
  protecting against.
- `main` and a PR both pass with the new value on every platform in the matrix, `windows-latest`
  included.

Related: atqamz/hand#416, atqamz/hand#413

atqamz:
## Summary

- Use `journal_mode=MEMORY` and `synchronous=OFF` only for writable SQLite connections in `-tags=test` builds, covering `internal/store` and `internal/registry`.
- Keep production helpers empty and read-only paths unchanged, preserving production rollback-journal/FULL durability.
- Set the unit-test timeout to 30m in `Makefile` and `.github/workflows/ci.yaml`; e2e and contract targets are unchanged.

Fixes atqamz/hand#416
Fixes atqamz/hand#430

## Test plan

- Red-green targeted tests verified test builds select `journal_mode=MEMORY` and `synchronous=OFF`.
- `nix develop -c make test` passed all packages with race detection and `-timeout=30m`; isolated `SECONDHAND_HOME` was used.
- Production build and default-tag `internal/store`/`internal/registry` tests passed.
- `nix develop -c make lint` passed with 0 issues.
- Linux before/after package durations (`go test -tags=test -race -count=1`, isolated `SECONDHAND_HOME`):

  | package | before | after |
  | --- | ---: | ---: |
  | `cmd` | 124.171s | 77.896s |
  | `internal/runtime` | 105.436s | 51.655s |
  | `internal/watcher` | 46.655s | 33.101s |
  | `internal/store` | 24.295s | 10.694s |

- Windows could not be measured locally; PR CI passed `Test (windows-latest)` in 7m13s, and all five platform unit checks plus e2e, contract, lint, and build checks passed.
- Production WAL is deferred to atqamz/hand#433 for separate crash/recovery, locking, Windows, and concurrent benchmark evidence.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -55,7 +55,8 @@ jobs:
         with:
           go-version-file: go.mod
       - name: Test
-        run: go test -tags=test -race ./...
+        # 30m covers the measured 4x Windows runner variance while keeping a real hang actionable.
+        run: go test -tags=test -race -timeout=30m ./...
       - name: Build
         run: go build -o hand .
 
--- a/Makefile
+++ b/Makefile
@@ -10,7 +10,8 @@ build:
 	CGO_ENABLED=0 go build -ldflags "$(LDFLAGS)" .
 
 test:
-	SECONDHAND_HOME=$$(mktemp -d) go test -tags=test -race ./...
+	# 30m covers the measured 4x Windows runner variance while keeping a real hang actionable.
+	SECONDHAND_HOME=$$(mktemp -d) go test -tags=test -race -timeout=30m ./...
 
 fmt:
 	gofmt -w .
--- a/internal/registry/registry.go
+++ b/internal/registry/registry.go
@@ -177,7 +177,7 @@ func OpenReadOnlyAt(path string) (*Registry, error) {
 }
 
 func open(path string, readOnly bool) (*sql.DB, error) {
-	query := "?_pragma=busy_timeout(5000)&_pragma=foreign_keys(1)&_txlock=immediate"
+	query := "?_pragma=busy_timeout(5000)&_pragma=foreign_keys(1)&_txlock=immediate" + sqliteTestPragmas
 	if readOnly {
 		query = "?mode=ro&_pragma=busy_timeout(5000)&_pragma=foreign_keys(1)&_pragma=query_only(1)"
 	}
--- a/internal/registry/sqlite_pragmas_prod.go
+++ b/internal/registry/sqlite_pragmas_prod.go
@@ -1,0 +1,5 @@
+//go:build !test
+
+package registry
+
+const sqliteTestPragmas = ""
--- a/internal/registry/sqlite_pragmas_testmode.go
+++ b/internal/registry/sqlite_pragmas_testmode.go
@@ -1,0 +1,5 @@
+//go:build test
+
+package registry
+
+const sqliteTestPragmas = "&_pragma=journal_mode(MEMORY)&_pragma=synchronous(OFF)"
--- a/internal/registry/sqlite_testmode_test.go
+++ b/internal/registry/sqlite_testmode_test.go
@@ -1,0 +1,32 @@
+//go:build test
+
+package registry
+
+import (
+	"path/filepath"
+	"testing"
+)
+
+func TestOpenUsesCheapDurabilityInTestBuild(t *testing.T) {
+	db, err := open(filepath.Join(t.TempDir(), "registry.db"), false)
+	if err != nil {
+		t.Fatal(err)
+	}
+	t.Cleanup(func() { _ = db.Close() })
+
+	var journalMode string
+	if err := db.QueryRow("PRAGMA journal_mode").Scan(&journalMode); err != nil {
+		t.Fatal(err)
+	}
+	if journalMode != "memory" {
+		t.Fatalf("journal mode = %q, want memory", journalMode)
+	}
+
+	var synchronous int
+	if err := db.QueryRow("PRAGMA synchronous").Scan(&synchronous); err != nil {
+		t.Fatal(err)
+	}
+	if synchronous != 0 {
+		t.Fatalf("synchronous = %d, want 0", synchronous)
+	}
+}
--- a/internal/store/sqlite_pragmas_prod.go
+++ b/internal/store/sqlite_pragmas_prod.go
@@ -1,0 +1,5 @@
+//go:build !test
+
+package store
+
+const sqliteTestPragmas = ""
--- a/internal/store/sqlite_pragmas_testmode.go
+++ b/internal/store/sqlite_pragmas_testmode.go
@@ -1,0 +1,5 @@
+//go:build test
+
+package store
+
+const sqliteTestPragmas = "&_pragma=journal_mode(MEMORY)&_pragma=synchronous(OFF)"
--- a/internal/store/sqlite_testmode_test.go
+++ b/internal/store/sqlite_testmode_test.go
@@ -1,0 +1,25 @@
+//go:build test
+
+package store
+
+import "testing"
+
+func TestOpenUsesCheapDurabilityInTestBuild(t *testing.T) {
+	db, _ := openTemp(t)
+
+	var journalMode string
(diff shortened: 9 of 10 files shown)


atqamz:
Merging. The acceptance criterion I cared most about is met in a stronger form than I asked for.

I asked for "a reader can follow from the code that a real `hand` invocation cannot reach the cheaper setting". A build-tag-gated constant is better than a runtime guard, because the cheap pragmas are not merely unreachable in a production binary, they are not compiled into it at all. Verified both build paths myself rather than taking the claim:

- `flake.nix` sets `checkFlags = [ "-tags=test" ]` and no `tags` attribute, so `buildGoModule`'s build phase compiles `sqliteTestPragmas = ""`.
- `.github/workflows/release.yaml` builds both the unix and Windows artifacts with `go build -ldflags ...` and no `-tags` at all.

Also worth stating for the record, since nothing in the PR says it and someone will eventually worry about it: `journal_mode=MEMORY` set per-connection is not persisted in the database file - only WAL is - so a test run cannot leave a database in a mode a later production open would inherit. That is why this seam is safe in a way a persisted mode would not be.

The measured Linux improvement is real and roughly halves the two packages that were panicking: `cmd` 124.2s to 77.9s, `internal/runtime` 105.4s to 51.7s, `internal/watcher` 46.7s to 33.1s, `internal/store` 24.3s to 10.7s.

One caveat I want recorded rather than glossed. `Test (windows-latest)` passing in 7m13s here is **one sample on one runner**, and runner variance is the whole subject of atqamz/hand#416: the same code measured 41.8s and 167.7s for `internal/store` on two different Windows runners hours apart. So this is evidence the fix helps, not proof the leg is stable. What makes me comfortable merging anyway is that the two mechanisms are independent - the 30m timeout absorbs the variance even on a bad runner, and the pragma change reduces how much variance there is to absorb. The three timing assertions in `internal/herdr` and `internal/watcher` that failed on `main` from I/O starvation passed here, which is the outcome the brief asked to be reported either way.

`main` at `53e59f5` is red on Windows right now, so this also restores it. Deferring production WAL to atqamz/hand#433 rather than folding it in was the right call and is exactly what the brief asked for; I have milestoned it 0.8.0.

atqamz:
Post-merge verification, in the shipped artifact rather than in the diff.

Rebuilt the supervisor's own binary at `ee46064d` with `make build` (`CGO_ENABLED=0 go build -ldflags "-s -w ..." .`, no `-tags`) and inspected it:

```
$ strings hand | grep -c 'journal_mode(MEMORY)'   -> 0
$ strings hand | grep -c 'synchronous(OFF)'       -> 0
$ strings hand | grep -c 'busy_timeout(5000)'     -> 2
```

Two `busy_timeout(5000)` for `internal/store` and `internal/registry`, and neither cheap pragma present anywhere in the binary. So the durability claim is not merely a code-review argument about build tags - the strings are absent from the artifact a user would run.

That command is worth keeping as the check for this seam: if a future change reintroduces the cheap setting into a production build, `strings` on the binary catches it without needing to reason about tag propagation.

`main` is green again as of this commit: run 33062900684 concluded success on every job. Windows went from two packages panicking at the 600s cap to `cmd` 272.3s, `internal/runtime` 142.1s, `internal/watcher` 75.9s, whole job 7m18s against 9m59s on the last green run before it.

PR #447, opened 2026-08-28
fix(attention): surface a not-submitted send to the supervisor

### LINKED ISSUE #419: a composer-timeout send leaves no record at all, so a worker the supervisor cannot reach produces no signal anywhere
## What happens

When `hand send` gives up on a busy composer it records the send as `not-submitted` and then never
mentions it again. Nothing surfaces it: not `hand status`, not `hand orient`, not `next_action`.

`state.SendNotSubmitted` is a first-class send state (`internal/state/types.go:57`), but
`taskAttentionEvidence` in `cmd/statusview.go` maps only three of them into attention evidence:

```go
if v.latestSend != nil {
    evidence.SendPending = v.latestSend.State == state.SendPending
    evidence.SendUncertain = v.latestSend.State == state.SendUncertain
    evidence.SendPartial = isSendPartial(v)
}
```

`SendNotSubmitted` produces no `attention.Subject`, so it reaches none of `classifyNextAction`'s
fourteen ranking branches (which cover `KindSendUncertain` and `KindSendPartial` but not this) and it
appears in no orientation surface. The failure is durably written down and then structurally invisible.

## Reproduction, four times on one task today

`403-cross-fleet-worktree-resolution`'s worker was mid-turn continuously while I tried to deliver
review feedback. Four consecutive attempts, `--wait` raised from the 2m default to 8m and then 10m:

```
error: "composer stayed busy for 10m0s; no external message mutation occurred (agent_status=\"working\")"
send_state: not-submitted
```

Each of those is a supervisor decision that never reached the worker. After each one, `hand orient`
reported nothing about it - the task showed `working`/`working` with no flag - and `hand status` showed
no send condition. If I had not been holding the fact in my own head and posting the same text as a
GitHub PR comment as a fallback, four pieces of steering would have been silently lost.

That is the failure `hand orient` exists to prevent. Remembering that I tried to steer a worker and
failed is exactly the state-reconstruction-from-memory that `AGENTS.md` forbids.

## Why the workaround is not one

There is no `hand` mechanism for deferred delivery. `--force` sends into a busy composer without
waiting, which is a different and riskier thing - text typed mid-turn is not the same as text
delivered when the worker is ready - and it is not a retry. `SendUndeliveredMessage` /
`SendUndeliveredAt` exist on the attempt record but are legacy fields, read only by
`internal/store/migrate.go` during import and never written or acted on by current code.

So the supervisor's only real fallback is out-of-band: post the message somewhere the worker happens
to read, like its PR. That works when there is a PR and the worker checks it. It is not a mechanism.

## Fix direction

The first item is the small one and it is the one that matters.

1. **Surface it.** Add a `SendNotSubmitted` case to `taskAttentionEvidence` and a corresponding
   attention kind, and rank it in `classifyNextAction` next to `KindSendUncertain`. The reason line
   should name what happened and what to do: "a send to this worker was never submitted
   (composer-timeout); resend or use --force". A supervisor must not have to remember this.
2. **Consider deferred delivery.** A recorded `not-submitted` send is a message hand already holds. It
   could be re-offered on the next `hand watch` event that observes the composer free, or on the next
   `hand orient`, rather than requiring the operator to reconstruct and retype it. If that is too
   much for one change, at least keep the message text retrievable so a retry does not depend on the
   operator still having it.
3. While in there: the error message says the composer "stayed busy for 10m0s ... (agent_status=
   \"working\")". `agent_status="working"` is genuinely useful - it distinguishes a busy worker from a
   wedged one. Say which of those the operator is looking at, since the action differs.

## Acceptance

- A `not-submitted` send makes the task actionable in `hand status` and `hand orient`, with a reason
  naming the cause.
- `classifyNextAction` can nominate it.
- The recorded message text remains retrievable for a retry.
- A test covers a `not-submitted` send producing an attention subject.

atqamz:
## Summary

- `state.SendNotSubmitted` was already durably recorded, but `taskAttentionEvidence` mapped only pending, uncertain, and the one not-submitted reason code (`enter-rejected-after-text-staged`) that `isSendPartial` already claimed as send-partial. A send rejected before the composer ever accepted text (`text-rejected-before-acceptance`) raised no attention subject, no status flag, and no next action.
- Adds `attention.KindSendNotSubmitted`, wired through `taskAttentionEvidence`/`taskFlags` in `cmd/statusview.go`, and ranked in `classifyNextAction` (`cmd/nextaction.go`) directly below `send-partial`: both are known negatives (unlike `send-uncertain`, hand knows the message did not land), but partial leaves residual text in the composer a blind retry could duplicate, while this case is the store's own retry-safe rejection.
- `internal/store.SendNeedsAttention` (the narrower per-send predicate `cmd/status.go`'s JSON `LatestSend.NeedsAttention` field reads directly) is untouched and keeps its existing answer for this reason code; only the task-level attention derivation changes, per `docs/adr/attention-is-one-derivation-over-three-channels.md`.

## Test plan

```
$ make lint
go vet ./...
golangci-lint run
0 issues.
go run ./tools/commentlint .

$ make test
ok  	github.com/atqamz/hand/cmd	93.681s
ok  	github.com/atqamz/hand/internal/attention	1.065s
... (all packages ok)

$ make e2e
ok  	github.com/atqamz/hand/tests/e2e	61.651s
```

New/changed tests:
- `cmd/nextaction_test.go`: `send-not-submitted outranks queued work` (the failing test written before the fix, confirming the issue's diagnosis) and `send-partial outranks send-not-submitted` (ranking).
- `internal/attention/attention_test.go`: `TestDeriveRaisesSendNotSubmittedDistinctFromSendUncertain`.
- `cmd/status_test.go`: updated `TestStatusFlagsPartialSendAfterFreshRead`'s "retry-safe rejection" case, now expecting the `send-not-submitted` flag and task-level attention, while splitting out the unchanged `internal/store.SendNeedsAttention`-backed JSON expectation into its own field so the two predicates can legitimately diverge for this reason code.
- `docs/testing-invariants.md`: added `INV-REP-6` citing the tests above.

Closes atqamz/hand#419

### DIFF
--- a/cmd/nextaction.go
+++ b/cmd/nextaction.go
@@ -20,6 +20,7 @@ const (
 	nextActionNeedsRepair      = "needs-repair"
 	nextActionSendUncertain    = "send-uncertain"
 	nextActionSendPartial      = "send-partial"
+	nextActionSendNotSubmitted = "send-not-submitted"
 	nextActionReportUnreadable = "report-unreadable"
 	nextActionUnreported       = "unreported"
 	nextActionRuntimeUnknown   = "runtime-unknown"
@@ -55,6 +56,10 @@ func classifyNextAction(cfg workerConfig, projectCount int, backlog backlogSumma
 	if v, ok := firstView(sorted, func(v taskView) bool { return hasAttentionKind(v, attention.KindSendPartial) }); ok {
 		return nextAction{Kind: nextActionSendPartial, Task: v.task.ID, Command: statusCommand(v.task.ID),
 			Reason: statusReason(v.task.ID, "confirm whether its send needs to be retried")}
+	}
+	if v, ok := firstView(sorted, func(v taskView) bool { return hasAttentionKind(v, attention.KindSendNotSubmitted) }); ok {
+		return nextAction{Kind: nextActionSendNotSubmitted, Task: v.task.ID, Command: statusCommand(v.task.ID),
+			Reason: statusReason(v.task.ID, "resend the message that was not submitted")}
 	}
 	if v, ok := firstView(sorted, func(v taskView) bool { return hasAttentionKind(v, attention.KindReportDecision) }); ok {
 		return nextAction{Kind: state.ReportNeedsDecision, Task: v.task.ID, Command: statusCommand(v.task.ID),
--- a/cmd/nextaction_test.go
+++ b/cmd/nextaction_test.go
@@ -22,6 +22,7 @@ func TestClassifyNextActionExactPrecedence(t *testing.T) {
 	queuedBacklog := backlogSummary{Queued: 1}
 	uncertainSend := &state.SendAttempt{State: state.SendUncertain}
 	partialSend := &state.SendAttempt{State: state.SendNotSubmitted, ReasonCode: state.SendReasonEnterRejectedAfterTextStaged}
+	notSubmittedSend := &state.SendAttempt{State: state.SendNotSubmitted, ReasonCode: state.SendReasonTextRejectedBeforeAcceptance}
 	pendingSend := &state.SendAttempt{State: state.SendPending}
 
 	tests := []struct {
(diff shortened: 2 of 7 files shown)


atqamz:
Review from the supervising session. One change needed before merge.

Code and tests look right, and two judgment calls here are better than what the brief asked for:

- `evidence.SendNotSubmitted = ... && !evidence.SendPartial` keeps partial and not-submitted from double-raising, so they stay distinguishable in every surface.
- The `status_test.go` change is not a silent assertion flip. Splitting `wantAttention` from `wantSendAttention` and naming why - task-level attention versus the narrower per-send `store.SendNeedsAttention` - documents a real distinction the old single field was hiding.

The change needed is in `docs/testing-invariants.md`. INV-REP-6 is overstated:

> Every send state hand durably records reaches at least one supervisor-visible attention subject, distinguishable from every other send state's.

`SendSubmitted` is durably recorded and must raise nothing - this PR's own `status_test.go` case asserts `submitted → wantAttention: false`. So as written the invariant is falsified by correct behavior, and the map is the wrong place for that: `docs/adr/tests-state-invariants-first-examples-second.md` accepts that a wrong invariant is worse than a missing one, because it reads as a specification and would license someone to "fix" submitted into raising attention. The cited coverage is also one case rather than the universal claim.

Please narrow it to what is true and actually checked, along these lines:

> A send state that records a delivery failure - pending, uncertain, partial, or not-submitted - reaches at least one supervisor-visible attention subject, and no two of them collapse to the same kind. A submitted send raises none.

Falsifiable, matches the three tests you wrote, and states the negative explicitly so the exception is not left implicit. Keep the same id, and cite the same tests plus the `submitted` case from `TestStatusFlagsPartialSendAfterFreshRead`.

Nothing else blocks. No code change needed.

PR #449, opened 2026-08-28
fix(harness): state ship/scout delivery authorization at launch

### LINKED ISSUE #411: nothing tells a ship worker it may commit, push, or open the PR, so a finished ship task cannot deliver
## What happens

A `kind: ship` task's deliverable is a pull request. Nothing at launch tells the worker it may commit,
push, or open one, and workers correctly refuse to do any of it without being told - so a ship task
whose brief forgot the grant finishes its work and then stops, holding a complete change it is not
allowed to deliver. `hand spawn` accepts such a brief, and `hand doctor` says nothing.

This is the same shape as atqamz/hand#406, one field over: hand knows the fact (`kind: ship`), the
worker needs the fact, and the only thing carrying it is supervisor prose that can silently omit it.

## Live evidence, two tasks, right now

`6-adoption-docs` and `6-consumer-fixture`, both `kind: ship`, project `omanixy`, both with live
harness processes. Neither brief contains any authorization language:

```
$ grep -il 'authoriz\|commit, push\|open the pull request' \
    data/6-adoption-docs/brief.md data/6-consumer-fixture/brief.md
                                        # no match in either
$ hand status 6-adoption-docs
kind: ship   state: done      pr: unknown
$ hand status 6-consumer-fixture
kind: ship   state: working   pr: unknown
```

Their own reports, once they were finally given a report channel (#406), say exactly what happened:

> `done:` README.md contains the adoption guide covering scope, prerequisites, installation, module
> ownership, feature and optional-integration status, state ownership, diagnostics, pinning, and
> attribution; ... only README.md is modified; **no merge or issue close performed**

> `working:` consumer fixture implementation is complete in the worktree: added `examples/consumer`
> with standalone and integrated NixOS/Home Manager configs, pinned local Omanixy input, fixture
> lock, root consumer-fixture checks, service/config ownership reuse, closure and independence
> assertions; just fmt, standalone fixture check, and canonical just check passed; **nothing remains
> technically, I am not blocked, and merge/issue closure are supervisor actions**

An earlier entry from the first is even more explicit: "no commit, push, merge, or issue close
performed". Two finished ship tasks, no commit, no branch, no PR, and `pr: unknown` on both. The work
exists only as uncommitted files in a worktree.

## Why prose is not enough

The worker behaviour here is *correct* and should not change - refusing to push without authorization
is the safe default, and it is what the global agent rules require. The defect is that hand lets a
ship task launch without ever supplying the authorization it structurally requires, so the safe
default silently becomes a dead end. A supervisor reading `state: done` has no signal distinguishing
"delivered" from "finished and mute and unable to deliver".

## Fix direction

1. **Inject it at launch, next to the report channel.** `internal/harness/harness.go`'s `briefPrompt`
   is already the place hand states launch-level facts, and atqamz/hand#406 adds the report path there.
   For `kind: ship`, state that the worker is authorized to commit, push its branch, and open the pull
   request, and that merge and issue closure are the supervisor's action only. For `kind: scout`, state
   that the deliverable is a report and that it must not push. hand already knows the kind - it prints
   it in `hand status`.
2. **Make `hand doctor` warn** for an open ship task with no recorded PR whose brief contains no
   authorization language, the same way #406's check warns on a missing report path.
3. Do not weaken the worker-side default. Nothing here should make a worker push when it was not told
   it may.

## Acceptance

- A ship task spawned from a brief with no authorization language still receives the grant at launch.
- A scout task receives the opposite statement and no grant.
- `hand doctor` names open ship tasks that cannot deliver.
- `make test` and `make lint` clean.

Related: atqamz/hand#406

atqamz:
## Summary

- `briefPrompt` now states, unconditionally at launch, that a `kind: ship` worker is authorized to commit, push its branch, and open the pull request (merge and issue closure stay the supervisor's action), and that a `kind: scout` worker's deliverable is a report and it must not push.
- Threaded the task kind through every `harness.Options` construction site (`spawn`, `reopen`, `promote`, `reconcile`) so the grant no longer depends on brief prose.
- `hand doctor` now warns on an open ship task that reported `done:` with no pull request recorded, naming `hand status <id>` as the way to see what unblocks it.

Closes atqamz/hand#411

## Test plan

```
$ make lint
go vet ./...
golangci-lint run
0 issues.
go run ./tools/commentlint .

$ make test
...
ok  	github.com/atqamz/hand/cmd	214.018s
...
ok  	github.com/atqamz/hand/internal/harness	3.296s
...
ok  	github.com/atqamz/hand/internal/runtime	172.851s
...
(all packages ok)

$ make e2e
ok  	github.com/atqamz/hand/tests/e2e	89.856s
```

Rendered launch prompt, real render via `harness.Build`, not source:

Ship: `...You are authorized to commit, push your branch, and open the pull request; merging and closing the issue are the supervisor's action only....`

Scout: `...Your deliverable is a report; you must not commit, push, or open a pull request....`

### DIFF
--- a/cmd/doctor.go
+++ b/cmd/doctor.go
@@ -242,6 +242,23 @@ func doctorFindings(fleetHome string, histories []state.TaskHistory) ([]doctorFi
 		}
 	}
 
+	for _, history := range histories {
+		task := history.Task
+		if task.Kind != state.KindShip || task.PR != "" {
+			continue
+		}
+		lines, err := state.ReadReportLines(fleetHome, task.ID)
+		if err != nil {
+			findings = append(findings, doctorFinding{Severity: doctorWarning, Text: fmt.Sprintf("task %q report could not be read: %v", task.ID, err)})
+			continue
+		}
+		last, ok := state.LastReportedState(lines)
+		if !ok || last.State != state.ReportDone {
+			continue
+		}
+		findings = append(findings, doctorFinding{Severity: doctorWarning, Text: fmt.Sprintf("task %q is an open ship task that reported done with no pull request recorded; run `hand status %s` to see what unblocks it", task.ID, task.ID)})
+	}
+
 	projects, err := project.ListReadOnly(fleetHome)
 	if err != nil {
 		return nil, err
--- a/cmd/doctor_test.go
+++ b/cmd/doctor_test.go
@@ -217,6 +217,59 @@ func TestDoctorWarnsForOpenBriefWithoutReportPath(t *testing.T) {
 	}
 }
 
+func TestDoctorWarnsForOpenShipTaskReportedDoneWithNoPR(t *testing.T) {
+	home := t.TempDir()
+	t.Chdir(home)
+	mkFleetDirs(t, home)
+	if _, err := agentsmd.Refresh(home); err != nil {
+		t.Fatal(err)
+	}
+	if _, err := skill.Refresh(home); err != nil {
+		t.Fatal(err)
+	}
+	mustConfigSet(t, settingHarness, harness.Claude)
+
+	for _, task := range []state.Task{
+		{ID: "stuck", Project: "demo", Kind: state.KindShip, Brief: "data/stuck/brief.md", Lifecycle: state.TaskOpen},
+		{ID: "delivered", Project: "demo", Kind: state.KindShip, Brief: "data/delivered/brief.md", Lifecycle: state.TaskOpen, PR: "https://github.com/atqamz/hand/pull/1"},
+		{ID: "working", Project: "demo", Kind: state.KindShip, Brief: "data/working/brief.md", Lifecycle: state.TaskOpen},
+		{ID: "scouting", Project: "demo", Kind: state.KindScout, Brief: "data/scouting/brief.md", Lifecycle: state.TaskOpen},
+	} {
+		if err := state.CreateTask(home, task); err != nil {
+			t.Fatal(err)
+		}
+	}
+	for id, report := range map[string]string{
+		"stuck":     "done: nothing left to do\n",
+		"delivered": "done: PR opened\n",
+		"working":   "working: still going\n",
+		"scouting":  "done: report written\n",
+	} {
+		if err := os.WriteFile(state.ReportPath(home, id), []byte(report), 0o644); err != nil {
+			t.Fatal(err)
+		}
+	}
+
+	histories, err := state.ListOpenHistoriesReadOnly(home)
+	if err != nil {
+		t.Fatal(err)
+	}
+	findings, err := doctorFindings(".", histories)
+	if err != nil {
+		t.Fatal(err)
+	}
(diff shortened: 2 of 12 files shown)


atqamz:
Reviewed. The design is right and the two-layer defence - state it at launch, and have doctor name the tasks that could not deliver - is what the issue asked for. One change before this lands.

**The wiring is untested.** Five runtime files change and none of them gains a test. `spawn.go`, `reopen.go`, `promote.go` and both `reconcile.go` arms each pass `taskKind`, and deleting any one of those assignments leaves every test in this PR green. `TestBuildShipCarriesDeliveryAuthorization` proves `briefPrompt` renders the grant when it is given a kind; nothing proves a kind ever arrives.

That is the same defect class as the issue itself: the fact exists, and nothing checks that it reaches the worker.

`r.deps.buildHarness` is already the seam for this. `internal/runtime/routing_test.go:487` and `reconcile_test.go:307` both capture the `harness.Options` a launch was built with. Add tests in that shape asserting `Options.Kind`:

- spawn of a ship task builds with `state.KindShip`
- spawn of a scout task builds with `state.KindScout`
- reopen carries the task's kind
- promote carries `state.KindShip`
- the `reconciliationActionConfirmLaunch` arm carries the task's kind

Four or five short cases against the existing seam. Then add the corresponding row to `docs/testing-invariants.md` - INV-AUTH-1 and INV-AUTH-2 as written are about the renderer, and the invariant that actually matters is that every launch path supplies a kind.

Two smaller notes, neither blocking:

- `TestBuildWithoutKindCarriesNoAuthorizationStatement` enshrines empty-kind-is-silent as supported behaviour. That is defensible given the doctor net, but it is the one place a future regression can hide, which is why the wiring tests above are the ones I want.
- The doctor check fires only on a last report of `done`. The issue's second live example (`6-consumer-fixture`) reported `working:` while saying "nothing remains technically", so this check would have missed it. That is fine - "working but actually finished" is not mechanically detectable - but say so in the invariant row rather than implying full coverage.

atqamz:
Addressed in c51a5a3.

Added wiring tests at the r.deps.buildHarness seam asserting Options.Kind for:
- spawn ship, spawn scout (`TestSpawnBuildsHarnessCarryingTaskKind`)
- reopen (`TestReopenBuildsHarnessCarryingTaskKind`)
- promote (`TestPromoteBuildsHarnessCarryingShipKind`)
- reconcile's confirm-launch arm (`TestReconcileConfirmLaunchCarriesTaskKind`)

Verified each one actually catches the defect class you named: deleted spawn.go's `taskKind: kind`
locally and confirmed `TestSpawnBuildsHarnessCarryingTaskKind` fails, then restored it.

Split the invariant map: INV-AUTH-1/2 are now scoped to the renderer given a kind, and a new
INV-AUTH-4 states the wiring invariant (every launch path supplies the kind). INV-AUTH-3's row now
notes the doctor check only fires on a last-reported `done`, not a worker still saying `working:`
about work it considers finished (accepted limit, not full coverage).

make lint and make test both pass. Not merging.

PR #450, opened 2026-08-28
feat(harness): catalogue codex's usage-limit signature

### LINKED ISSUE #435: watcher: codex has no usage-limit signature, so a credit-exhausted worker reads as stale and idle-unreported instead of held on limit
## What happens

`internal/harness/usagelimit.go` catalogues a usage-limit signature for `claude` only, and says so
deliberately:

> Only claude has a signature; every other harness declines the capability until someone catalogues
> its refusal against a real limited run, the bar `firstRunPrompts` holds.

This is that limited run. Two codex workers in this fleet hit the limit simultaneously on
2026-08-27 at roughly 18:00 WIB, and `hand watch` classified both as `stale`, then
`parked`, then `idle-unreported` - never `limit`.

## The verbatim refusal, and why the claude regex cannot match it

Read from both panes, identical text:

```
■ You've hit your usage limit. Upgrade to Pro (https://chatgpt.com/explore/pro), visit https://chatgpt.com/codex/settings/usage to purchase more credits or try again at 7:27 PM.
```

The catalogued `claude` pattern is anchored on **reached**:

```go
Match: regexp.MustCompile(`(?i)usage limit reached|\d+-hour limit reached|reached your (?:\w+ )*limit`)
```

codex says **hit**, not **reached**, so it cannot match, and the anchoring comment - "Anchored on
*reached*, never `limit` alone, so an approaching-your-limit warning cannot read as a stop" - is the
reason it should not be loosened to bare `limit` either. codex needs its own entry.

## The reset format is the awkward part

codex reports `try again at 7:27 PM`. Unlike claude's `usage limit reached|<epoch>`, this is:

- **local** time, not UTC, and it names no zone. Observed at 18:21 WIB (`+0700`) predicting 19:27 WIB.
- **12-hour** with an uppercase meridiem and no leading zero.
- **dateless**, so a refusal seen at 23:50 predicting `12:15 AM` means the next day. A parser has to
  resolve against `now` in the local zone and roll the date forward when the parsed clock time is
  earlier than `now`.

Per the existing doc comment, `Reset` is only ever a prediction and nothing decides the limit is over
from it, so a wrong parse degrades to a late retry rather than a wrong resume. Returning
`false` when the wording is anything but this exact shape is fine.

## What it cost, which is the argument for fixing it

`hand orient` reported `next_action_kind: unreported` with
`next_action_reason: "it stopped without reporting a terminal state"`, and `hand status --fields
id,state,held,flags` showed `unreported parked` for both tasks. That is indistinguishable from a
wedged or crashed worker, so I investigated both as if they were - checked the report file, the gate
pipeline state, `agent_status`, and sent two steering messages that could not be consumed - before
reading enough pane scrollback to find the refusal. A `limit` hold would have said "resumes at 19:27,
do nothing" immediately, which is exactly what the hold kind exists for: `hand hold set` rejects it
with "hold kind limit is set by hand watch when a worker's harness stops on a usage limit, and
cleared when it runs again", so the operator cannot record it manually either.

Second-order effect worth noting for whoever fixes this: a codex limit is **account-wide**, so it
stops every codex-routed worker in the fleet at once, plus any no-mistakes gate agent running on the
same account. A fleet whose projects all route to codex - which this one does for `hand` and
`omanixy` - goes from "several workers in flight" to "nothing can progress" with no signal beyond
per-task staleness. Whether that deserves a fleet-level rather than per-task surface is a design
question, not something this issue asks for.

## Acceptance

- A codex worker stopped on its usage limit is reported as held on `limit`, not as `stale`,
  `parked`, or `idle-unreported`.
- The hold carries the predicted reset instant when the refusal names one, and omits it rather than
  guessing when it does not.
- `try again at <h>:<mm> <AM|PM>` resolves in the local zone and rolls to the next day when the named
  clock time is earlier than now.
- The claude signature keeps matching everything it matches today; a test pins both wordings.
- Nothing decides the limit is over from the prediction alone, consistent with the existing contract
  in `internal/watcher`'s usagelimit handling.

## Diagnostics

`hand` at `ee46064d`, `distribution: source`. Panes `wK:pA` (`412-pool-slot-integrity`) and `wM:p9`
(`6-consumer-fixture`), both `agent: codex`, `model: gpt-5.6-luna`, `agent_status: done`.

Related: atqamz/hand#419, atqamz/hand#426

atqamz:
## Summary

- codex stops on a usage limit with "You've hit your usage limit ... try again at 7:27 PM" (atqamz/hand#435), which claude's `reached`-anchored pattern cannot match, so `hand watch` classified it `stale`, then `parked`, then `idle-unreported` instead of held on `limit`.
- Adds a codex entry anchored on "hit" - the word codex actually uses, for the same reason claude's is anchored on "reached", so an approaching-limit warning still cannot read as a stop.
- Adds a reset parser for codex's dateless, zoneless 12-hour clock wording (`try again at 7:27 PM`), resolved against `now` in the local zone with rollover across midnight. Any wording but the exact observed shape degrades to no prediction, never a guessed one.
- Extends `docs/testing-invariants.md` with the invariants this catalogue holds.

Closes atqamz/hand#435

## Note for the watcher, not fixed here

The issue observes that a codex usage limit is account-wide: it stops every codex-routed worker in the fleet at once. `internal/watcher/usagelimit.go` and `internal/runtime/liveness.go` both detect and schedule per attempt/pane, with no fleet-wide awareness. Nothing here makes that assumption more visible than it already was - the issue explicitly calls this a design question, not something this change should fix - but flagging it per the brief's ask.

## Test plan

```
$ make lint
go vet ./...
golangci-lint run
0 issues.
go run ./tools/commentlint .

$ make test
...
ok  	github.com/atqamz/hand/internal/harness	0.005s
...
ok  	github.com/atqamz/hand/internal/runtime	106.766s
...
ok  	github.com/atqamz/hand/internal/watcher	47.758s
...
(full suite green)

$ make e2e
ok  	github.com/atqamz/hand/tests/e2e	170.971s
```

New/updated tests: `TestSupportsUsageLimitOnlyWhereASignatureIsCatalogued`, `TestDetectUsageLimitDoesNotCrossMatchClaudeAndCodex`, `TestDetectUsageLimitRecognizesTheObservedCodexRefusal` (verbatim string from the issue), `TestDetectUsageLimitIgnoresCodexTextThatIsNotAStop`, `TestDetectCodexUsageLimitResetInstant` (observed shape, unparseable shape, midnight rollover, later-today clock time). `TestDetectUsageLimitRecognizesEveryCatalogedWording` and the rest of claude's existing suite pass unchanged, confirming claude's pattern still matches everything it matched before.

### DIFF
--- a/docs/testing-invariants.md
+++ b/docs/testing-invariants.md
@@ -81,11 +81,11 @@ Source: [Usage-limit detection is a harness capability](adr/usage-limit-detectio
 
 | id | invariant | layer | coverage |
 |---|---|---|---|
-| INV-LIMIT-1 | A catalogued harness's stop refusal is detected as a usage limit; an uncatalogued harness never is, for any text. | property | `TestSupportsUsageLimitOnlyWhereASignatureIsCatalogued`, `TestDetectUsageLimitDeclinesForAnUncataloguedHarness` |
+| INV-LIMIT-1 | A catalogued harness's stop refusal is detected as a usage limit; an uncatalogued harness never is, for any text. | property | `TestSupportsUsageLimitOnlyWhereASignatureIsCatalogued`, `TestDetectUsageLimitDeclinesForAnUncataloguedHarness` land unit cases; property form unaudited |
 | INV-LIMIT-2 | No catalogued signature matches an approaching-limit warning, only the stop itself. | unit | `TestDetectUsageLimitIgnoresTextThatIsNotAStop`, `TestDetectUsageLimitIgnoresCodexTextThatIsNotAStop` |
 | INV-LIMIT-3 | One harness's signature never matches another harness's wording. | unit | `TestDetectUsageLimitDoesNotCrossMatchClaudeAndCodex` |
-| INV-LIMIT-4 | A reset instant that does not parse degrades to no prediction, never to a guessed one. | unit | `TestDetectCodexUsageLimitResetInstant` |
-| INV-LIMIT-5 | A dateless clock-time reset resolves against `now` in the local zone, rolling forward when the named time is not after `now`. | unit | `TestDetectUsageLimitReadsTheResetInstant`, `TestDetectCodexUsageLimitResetInstant` |
+| INV-LIMIT-4 | A reset instant that does not parse degrades to no prediction, never to a guessed one, for any unrecognized wording. | property | `TestDetectCodexUsageLimitResetInstant` lands a unit case; property form unaudited |
+| INV-LIMIT-5 | A dateless clock-time reset resolves against `now` in the local zone, rolling forward when the named time is not after `now`, for any (now, clock time) pair. | property | `TestDetectUsageLimitReadsTheResetInstant`, `TestDetectCodexUsageLimitResetInstant` land unit cases; property form unaudited |
 
 ## Registry and fleet identity
 
--- a/docs/testing-invariants.md
+++ b/docs/testing-invariants.md
@@ -74,6 +74,19 @@ Source: [hand init is the canonical fleet reconciler](adr/hand-init-is-the-canon
 | INV-INIT-5 | A non-empty directory is adopted if and only if its `go.mod` declares hand's own module path. | property | phase 0 of atqamz/hand#436 landed unit tests; property form unaudited |
 | INV-INIT-6 | Preflight refusal is total: a refused target has no byte changed, including the private runtime root. | property | unaudited |
 
+## Usage-limit signatures
+
+Source: [Usage-limit detection is a harness capability](adr/usage-limit-detection-is-a-harness-capability.md),
+`internal/harness/usagelimit.go`.
+
+| id | invariant | layer | coverage |
+|---|---|---|---|
+| INV-LIMIT-1 | A catalogued harness's stop refusal is detected as a usage limit; an uncatalogued harness never is, for any text. | property | `TestSupportsUsageLimitOnlyWhereASignatureIsCatalogued`, `TestDetectUsageLimitDeclinesForAnUncataloguedHarness` |
+| INV-LIMIT-2 | No catalogued signature matches an approaching-limit warning, only the stop itself. | unit | `TestDetectUsageLimitIgnoresTextThatIsNotAStop`, `TestDetectUsageLimitIgnoresCodexTextThatIsNotAStop` |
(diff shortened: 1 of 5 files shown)


atqamz:
Reviewed. The signature, the parser and the tests are right, and two things I would have flagged you got ahead of: switching the two pre-existing tests off codex as the uncatalogued stand-in, and mirroring `parseClaudeReset`'s rollover rationale rather than inventing a second one. The `hit` anchoring with the reason stated next to claude's `reached` anchoring is exactly the right comment to leave behind.

One correction, then this can land.

**INV-LIMIT-1 claims a coverage it does not have.** As written:

```
| INV-LIMIT-1 | A catalogued harness's stop refusal is detected as a usage limit; an uncatalogued harness never is, for any text. | property | `TestSupportsUsageLimitOnlyWhereASignatureIsCatalogued`, `TestDetectUsageLimitDeclinesForAnUncataloguedHarness` |
```

"for any text" is a universal quantifier, so `property` is the right layer - but the two cited tests are table-driven unit tests over a handful of strings, and the coverage column reads as if the property is checked. The convention already in this file is to say what actually exists, e.g.:

```
| INV-INIT-5 | ... | property | phase 0 of atqamz/hand#436 landed unit tests; property form unaudited |
```

Match that: keep `property` as the layer, and have the coverage column name the unit tests **and** say the property form is unaudited. Same question for INV-LIMIT-4 and INV-LIMIT-5 - if either is a universal over inputs rather than the specific cases you enumerated, say so the same way.

This matters more than its size. This document is the spine of atqamz/hand#442, which will audit every row in it, and a row that overstates its own coverage is the one kind of entry that makes the whole map untrustworthy.

On the account-wide finding: agreed it is out of scope and correctly flagged rather than fixed. I am not filing it - each limited codex pane will show its own refusal and get its own hold, so the state stays correct and only the probing is redundant.

atqamz:
Fixed in 77e39b2: INV-LIMIT-1, INV-LIMIT-4, and INV-LIMIT-5 all restated as universals ("for any text", "for any unrecognized wording", "for any (now, clock time) pair"), so all three keep `property` as the layer with coverage naming the unit tests and stating the property form is unaudited, matching INV-INIT-5. INV-LIMIT-2 and INV-LIMIT-3 are enumerated-case claims, not universals, so left as `unit`.

Pushed, not merged.

PR #451, opened 2026-08-28
fix(watcher): name the recorded lock holder in a contended refusal

### LINKED ISSUE #410: lock error blames a nonexistent watcher when a long hand send holds the fleet-home lock
## What happens

`hand send --wait <long>` holds the fleet-home lock for the whole time it waits on a busy composer.
While it does, `hand watch --until-event` refuses - and blames a watcher that does not exist:

```
error: "a watcher is already attached to this fleet home - a watcher already holds the fleet-home
lock; stop it through its owning session, or use --takeover for cooperative replacement"
```

There is no watcher. The message names the wrong holder, and every remedy it offers is a dead end:
there is no owning session to stop, and `--takeover` would replace a watcher that is not there.

## Reproduction

```
$ hand send 406-brief-report-channel --file msg.txt --wait 8m &     # worker composer busy
$ pgrep -af 'hand watch'
                                                                    # nothing
$ hand watch --until-event
error: "a watcher is already attached to this fleet home - a watcher already holds the fleet-home lock; ..."
kind: precondition
exit: 3
```

Then, the moment the send returns:

```
$ hand send ... --wait 8m
send_state: submitted

$ hand watch --until-event
reported-done 406-brief-report-channel: implementation committed and pushed; pull request 409 is open
working 406-brief-report-channel: addressing review feedback by rejecting empty report paths in harness builds
exit=0
```

Same command, no other change, works immediately. The blocker was the send, not a watcher.

## Why it costs more than a confusing string

`AGENTS.md` requires `hand orient` at the start of every supervisor turn, and while this is happening
`hand orient` reports `monitor_state: unknown` with a next action of `monitor-rearm` - "run `hand watch
--until-event` to establish monitoring". So hand tells the supervisor to run a command, and that command
refuses with a diagnosis pointing at a stale lock that isn't stale. I went looking for a leaked watcher
process and found none, which is the only reason I did not go on to `--takeover` - and a `--takeover`
here would have been an action taken against a wrong diagnosis.

## Two asks

**1. Name the actual holder.** The lock error should report what holds the lock - command, and ideally
pid or start time - rather than assuming a watcher. `--takeover` should only be suggested when the
holder is in fact a watcher.

**2. Reconsider the lock's scope for `send`.** A send that is merely *waiting* for a composer to free
has not mutated anything yet. Holding the fleet-home lock across a wait that can legitimately run for
minutes blocks all monitoring for that whole window, and `--wait` defaults to 2m with longer values
being reasonable when a worker is genuinely busy. Either take the lock only around the parts that need
it (the ownership check and the recorded send), or make the wait interruptible by a watcher arming.
Whichever, say so explicitly in the implementation rather than leaving the coupling implicit.

## Acceptance

- With a long `hand send --wait` in flight, `hand watch --until-event` either succeeds or fails with a
  message naming the send as the holder.
- No error path suggests `--takeover` for a non-watcher holder.
- A test covers the send-holds-lock-then-watch-arms sequence.

atqamz:
## Summary

- `contend()` refused a busy `state/watch.pid.lock` by asserting "a watcher"
  regardless of what actually held it. The real holder is often an in-process
  supervision-wait / claude-stop / codex-stop bridge cycle: no owning session
  to stop, no takeover endpoint listening, invisible to `pgrep -af 'hand watch'`.
- `OwnerRecord` now carries a `Kind` (`watch`/`bridge`), set at acquisition by
  the new `AcquireBridgeContext` that `waitOwned` calls instead of
  `AcquireContext`. `contend()` reads the recorded owner before composing its
  message and matches the treatment to the holder it found: a bridge holder is
  named and told its hold is transient with no `--takeover` offered (and a
  `--takeover` attempt against one fails immediately instead of waiting out the
  grace period); an unreadable or absent record says so rather than guessing;
  anything else keeps the existing pid/generation-named remedy.
- Lock acquisition, release, ownership transfer, and the takeover protocol are
  unchanged - only what the refusal says and reads. Rationale and rejected
  alternatives are in `docs/adr/a-contended-refusal-names-its-recorded-holder.md`.

Closes atqamz/hand#410

## Test plan

```
$ make lint
go vet ./...
golangci-lint run
0 issues.
go run ./tools/commentlint .
$ make test
ok  	github.com/atqamz/hand/cmd	290.712s
...
ok  	github.com/atqamz/hand/internal/supervision	48.606s
ok  	github.com/atqamz/hand/internal/watcher	115.861s
... (all packages ok)
$ make e2e
ok  	github.com/atqamz/hand/tests/e2e	88.617s
```

New e2e test `TestWatchNamesALiveSupervisionBridgeHolderAndOffersNoTakeover`
reproduces atqamz/hand#410's exact live shape (issuecomment-5450929728) with a
real `hand supervision wait` subprocess holding the lock, then a contending
`hand watch --until-event` against it. Real output from that run:

```
$ hand watch --until-event --poll 30ms
exit 3
stderr: error: "a watcher is already attached to this fleet home - pid 1714540 holds the fleet-home watcher lock as a background supervision-wait cycle (generation 24776a7a3390db12118036d89c847e52), not an interactive watcher; it is transient and never honors --takeover - wait for its current cycle to end and retry"
kind: precondition
exit: 3
```

No owning-session instruction and no `--takeover` suggestion against this
holder, matching the ship brief's locked decisions.

### DIFF
--- a/docs/adr/a-contended-refusal-names-its-recorded-holder.md
+++ b/docs/adr/a-contended-refusal-names-its-recorded-holder.md
@@ -1,0 +1,98 @@
+# A contended refusal names its recorded holder
+
+- Date: 2026-08-28
+- Status: accepted
+- Issues: atqamz/hand#410
+- PRs: none
+
+## Context
+
+`contend` refused a busy `state/watch.pid.lock` with "a watcher is already attached to this fleet
+home", unconditionally, regardless of what actually held it. That was often wrong: `waitOwned`
+(`internal/supervision/wait.go`), reached from `hand supervision wait`, `claude-stop`, and
+`codex-stop`, acquires the exact same kernel lock in-process whenever no watcher currently owns the
+fleet home, to poll for actionable episodes on behalf of a Supervisor Harness's Stop hook. Its
+argv never contains "watch", so `pgrep -af 'hand watch'` finds nothing, and an operator who trusted
+the message went looking for a watcher process that did not exist.
+
+The refusal's remedy compounded the misdiagnosis. "Stop it through its owning session" named a
+session an in-process bridge cycle does not have, and "use --takeover for cooperative replacement"
+promised a protocol `waitOwned`'s holder cannot honor: `hand watch` wires `Ownership.
+TakeoverRequested()` into its own cancellation (`cmd/watch.go:110`), but `waitOwned` never reads
+that channel at all, so a takeover request against it is accepted by the endpoint and then ignored
+by the process that should act on it. A `--takeover` contender would wait out the full grace period
+and fail, having changed nothing.
+
+The information needed to fix this was already durable by the time contention is possible:
+`publishNewOwner` writes `state/watch.pid` and `state/watch.owner` while holding the authoritative
+lock, before any contender can observe it busy. `contend` simply never read them. Confirmed live on
+atqamz/hand#410 (issuecomment-5450929728): a real, alive `hand supervision claude-stop` process held
+the lock, and `state/watch.owner` named its pid and generation the entire time.
+
+Naming the pid was not previously done because an earlier decision
+([Watcher takeover is generation-attributed](watcher-takeover-is-generation-attributed.md)) removed
+pid as a *takeover target* - a stale or reused pid must never be signaled. That constraint is about
+the takeover protocol's mechanics, not about diagnostic text: reporting a pid does not signal it.
+
+## Decision
+
+`contend` reads the owner record before composing a refusal, and the message it builds depends on
+what that record shows:
+
+- A readable record naming a bridge holder (see below) says so, states that the hold is transient,
+  and does not offer `--takeover` - neither in the immediate refusal nor as a suggestion, because
+  it cannot succeed.
+- A readable record naming anything else keeps the existing remedy: stop it through its owning
+  session, or use `--takeover`.
+- An absent or unreadable record says exactly that - "the lock is held, but its ownership record
+  could not be read to identify the holder" - and offers no treatment it cannot back with evidence.
+  This satisfies [Every diagnosis names a reachable treatment](every-diagnosis-names-a-reachable-treatment.md):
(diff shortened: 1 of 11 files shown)


atqamz:

Three things this got right that were easy to get wrong:

**It confronted the ADR tension instead of stepping around it.** The old test asserted the refusal contains no `pid `, citing [Watcher takeover is generation-attributed](../blob/main/docs/adr/watcher-takeover-is-generation-attributed.md). Changing an assertion an ADR appears to back is exactly where a worker either quietly weakens an invariant or gives up on the fix. The new ADR reads that constraint correctly instead: it is about the takeover protocol's mechanics - a stale or reused pid must never be *signaled* - and reporting a pid in diagnostic text does not signal it. The older ADR's own rejected-alternatives list says "Signaling the advisory pid", not naming it.

**The unreadable-record branch offers no treatment.** `holderRecordUnreadable` says the lock is held and the record could not be read, and stops there. That is the harder half of [Every diagnosis names a reachable treatment](../blob/main/docs/adr/every-diagnosis-names-a-reachable-treatment.md) - an unnamed holder gets no unreachable promise either - and it would have been easy to fall back to the old wording as though it were known.

**Failing `--takeover` immediately against a recorded bridge holder** rather than waiting out `takeoverGrace` first. The grace period would only delay the same honest failure.

On `readOnlyMonitorState`: the decision not to change it, and the reasoning recorded in the report addendum, are right. The race is a timing window that cannot close without a coordination mechanism the brief's non-goals exclude, and what actually harmed the operator was the false diagnosis at the end of it, which this fixes. Deciding and recording beats silently skipping, which is what the brief asked for.

One observation rather than a request: `holderOther` covers both a watch-kind record and a legacy record with no `Kind`. Defaulting an unrecognized kind to "offer takeover" is the safe direction, since a takeover that cannot work fails loudly while a withheld one strands the operator. Worth keeping in mind if a third holder kind ever appears.

PR #453, opened 2026-08-28
fix(harness): append the report channel to grok's and pi's brief file

### LINKED ISSUE #418: grok and pi workers are structurally mute: neither builder carries a prompt, so #406's report channel never reaches them
## What happens

A worker launched on the `grok` or `pi` harness has no way to report anything. Both builders hand the
brief over as a file argument and pass no prompt at all:

```go
func buildGrok(o Options) launch.LaunchSpec {
	return launch.LaunchSpec{Executable: Grok, Args: []string{"--trust", "--file", o.Brief}}
}
func buildPi(o Options) launch.LaunchSpec {
	return launch.LaunchSpec{Executable: Pi, Args: []string{o.Brief}}
}
```

`briefPrompt` - the only place that tells a worker its report channel, after atqamz/hand#406 /
atqamz/hand#409 - is unreachable from either. `promptCapable` lists only claude, codex, opencode and
antigravity.

So the report-channel fix is complete for four of six harnesses, and a grok or pi worker is
structurally mute: it cannot append to `state/<id>.status` because nothing ever names that file to it,
and hand will classify it `parked` / `unreported` for the whole of its life. That is the exact failure
mode atqamz/hand#406 was filed for, and it survives atqamz/hand#409 for these two.

## Why the gap is currently untracked

`CarriesPrompt`'s doc comment says the missing prompt content "needs flags verified against a real
--help (atqamz/hand#36)". atqamz/hand#36 is **closed**, and it was about something else entirely -
verifying herdr's *agent detection* manifests against real codex, pi and grok panes. It says nothing
about prompt carriage. So the citation points a reader at a closed, unrelated issue and the real gap
has no home. That stale reference should be repointed here.

The consequence has also grown since that comment was written. It used to be that a grok or pi worker
missed `agentsmd.OperatorDecisionRule` and the front-matter disclaimer - bad, but the worker still
functioned. Now it also misses the report channel, which means it cannot communicate at all. The
comment understates it.

## Fix direction

Two independent pieces; the first is the one that matters.

1. **Give the file-based harnesses a prompt channel, or refuse to dispatch to them.** Either verify a
   real prompt flag against each CLI's `--help` and add it to `promptCapable`, or - if neither CLI has
   one - make `internal/runtime/dispatch.go` reject a route to a non-prompt-capable harness outright
   rather than launching a worker that cannot speak. Silently launching a mute worker is the worst of
   the three options and is what happens today.

   A third option worth weighing: for a file-based harness, **append the prompt text to the brief
   file itself** at provision time. The brief is already the one thing both CLIs read, `hand` already
   owns writing it, and the report-channel sentence plus the operator-decision rule are exactly the
   kind of content a brief can carry. That would make `CarriesPrompt` a statement about *how* the
   prompt is delivered rather than *whether* it is.

2. **Repoint the stale citation** in `CarriesPrompt`'s doc comment from atqamz/hand#36 to this issue,
   and state the current consequence: no report channel, not merely no operator-decision rule.

## Acceptance

- Either a grok/pi worker receives its report path and the operator-decision rule, or `hand spawn` /
  `hand reopen` refuses to route to grok/pi with a message naming this limitation.
- No code comment cites atqamz/hand#36 for prompt carriage.
- A test covers whichever behaviour is chosen, per harness.

Related: atqamz/hand#406, atqamz/hand#409

atqamz:
## Summary

- grok and pi took no prompt argument, so `briefPrompt`'s report path and operator-decision rule never reached them and hand launched a structurally mute worker anyway.
- Neither CLI has a verified prompt flag to add instead (not installed in this environment, and the standing rule is no unverified flag reaches `promptCapable`), so hand now appends the same `launchStatement` every other harness gets to the brief file itself at provision time, in a marked `---`-delimited appendix that reads as hand's text rather than the supervisor's brief. The append is idempotent so a reopen/resume does not grow a second copy.
- `CarriesPrompt` now means "receives the report path and operator-decision rule by some channel" rather than "takes a CLI prompt argument"; grok and pi both return `true`, and their doc comment is repointed from the closed, unrelated atqamz/hand#36 to atqamz/hand#418.
- Updated the downstream tests/warnings that assumed grok/pi could never carry the prompt (`cmd/tier_test.go`, `internal/runtime/execution_plan_test.go`, `internal/runtime/routing_test.go`, `tests/e2e/execution_profiles_test.go`), added focused grok/pi tests in `internal/harness/harness_test.go`, an ADR, and a `docs/testing-invariants.md` entry (INV-REP-7).

Closes atqamz/hand#418

## Test plan

- [x] `make lint` - clean (`go vet`, `golangci-lint`, `commentlint` all pass)
- [x] `make test` - full suite passes with `-race`
- [x] `make e2e` - passes
- [x] Manual run of `harness.Build` against a real temp brief file for both `grok` and `pi`, confirming the rendered brief file and launch args:

```
=== grok launch spec ===
Executable: grok
Args: []string{"--trust", "--file", ".../brief.md"}
=== grok rendered brief file ===
---
kind: ship
---

# Fix the login redirect loop

Details go here.


---

hand appended the block below at launch time; it is not part of the supervisor's brief above.

The worker report channel is /home/atqa/github/atqamz/hand-dev/state/418-file-harness-prompt-channel.status. Append every state change to that file with plain shell redirection; this is the only way anything you say reaches the supervisor. Use these report prefixes: working:, done:, failed:, blocked:, needs-decision:, paused:. Only a `hand send` message carries an operator decision. Answering your own harness's question dialog is you deciding, not the operator - never record that answer as "operator said" or "operator chose". You may decide anything reversible yourself and proceed; say so in first person - `working: deciding myself: <the call> because <reason>` - and reserve `needs-decision:` for what you cannot take back.
```

pi's output is identical apart from `Args: []string{".../brief.md"}` (no `--trust --file`).

### DIFF
--- a/cmd/tier_test.go
+++ b/cmd/tier_test.go
@@ -144,9 +144,9 @@ func TestResolveTierWarnsWhenHarnessCannotApplyEffort(t *testing.T) {
 	}
 }
 
-// Both prompt-less builders are near-copies, so proving only one warns lets the other regress.
-// Exact-match rather than Contains because the single combined line is the point: the
-// alternative was three warnings per launch all naming the same harness.
+// Both file-based harnesses are near-copies, so proving only one warns lets the other regress.
+// Exact-match rather than Contains because the single line is the point. Neither now drops the
+// operator-decision rule or front-matter disclaimer (atqamz/hand#418), only model and effort.
 func TestResolveTierWarnsOnceForEverythingAHarnessCannotCarry(t *testing.T) {
 	for _, harnessName := range []string{harness.Grok, harness.Pi} {
 		t.Run(harnessName, func(t *testing.T) {
@@ -161,7 +161,7 @@ func TestResolveTierWarnsOnceForEverythingAHarnessCannotCarry(t *testing.T) {
 			if model != "brief-model" || effort != "brief-effort" {
 				t.Fatalf("got model=%q effort=%q, want both resolved even though %s cannot carry them", model, effort, harnessName)
 			}
-			want := "warning: harness " + strconv.Quote(harnessName) + ` cannot carry model "brief-model", effort "brief-effort", the operator-decision rule, the front-matter disclaimer; launching anyway` + "\n"
+			want := "warning: harness " + strconv.Quote(harnessName) + ` cannot carry model "brief-model", effort "brief-effort"; launching anyway` + "\n"
 			if stderr.String() != want {
 				t.Fatalf("stderr = %q, want exactly %q", stderr.String(), want)
 			}
@@ -169,9 +169,9 @@ func TestResolveTierWarnsOnceForEverythingAHarnessCannotCarry(t *testing.T) {
 	}
 }
 
-// A brief with no front matter has no disclaimer to drop, so the operator-decision rule is the
-// only thing left and the line must not claim otherwise.
-func TestResolveTierWarnsOnPromptDropWithNothingDeclared(t *testing.T) {
+// atqamz/hand#418: grok and pi now receive the operator-decision rule by appending it to the
+// brief file, so a launch with nothing else declared produces no capability warning at all.
+func TestResolveTierNoWarningWhenFileBasedHarnessCarriesPrompt(t *testing.T) {
 	for _, harnessName := range []string{harness.Grok, harness.Pi} {
 		t.Run(harnessName, func(t *testing.T) {
 			home := t.TempDir()
@@ -181,9 +181,8 @@ func TestResolveTierWarnsOnPromptDropWithNothingDeclared(t *testing.T) {
 			if _, _, _, err := resolveTier(cmd, home, briefAbs, harnessName, "", ""); err != nil {
 				t.Fatal(err)
 			}
-			want := "warning: harness " + strconv.Quote(harnessName) + " cannot carry the operator-decision rule; launching anyway\n"
-			if stderr.String() != want {
-				t.Fatalf("stderr = %q, want exactly %q", stderr.String(), want)
+			if stderr.Len() != 0 {
+				t.Fatalf("stderr = %q, want no warning", stderr.String())
 			}
 		})
 	}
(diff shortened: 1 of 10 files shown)


atqamz:
Reviewed. The choice is right and well-argued: taking the brief-append option, refusing to guess a `--prompt` flag against a CLI you could not run, and keeping the refusal fallback documented in the ADR's rejected alternatives rather than discarding it. Extracting `launchStatement` so both channels share one wording is exactly the shape the issue wanted, and lifting the mechanical-class refusal for grok and pi follows from the fix rather than accommodating a regression - the six updated tests were asserting a consequence of the gap, not an invariant.

**One change before this lands: the side effect is in the wrong layer.**

`appendLaunchStatement` runs inside `buildGrok` and `buildPi`, so `harness.Build` now writes to the filesystem. It used to be a pure function from `Options` to a `LaunchSpec`. Your own ADR says the append happens "at provision time, before the launch command runs" - but `Build` is not only called at provision time.

`internal/runtime/reconcile.go:1377`, the `reconciliationActionConfirmLaunch` arm, calls `r.deps.buildHarness` to reconstruct *persisted launch evidence*. For a grok or pi attempt that now:

- writes to `data/<id>/brief.md` during what is supposed to be an observation, and
- fails outright if the brief file is gone, where it previously produced evidence regardless.

The marker check makes it idempotent, so nothing is corrupted today. But reconcile silently repairing a missing appendix is exactly the "converge by doing, not by observing" inversion the reconcile path is built to avoid, and a build that can fail on a missing file changes that arm's failure surface.

Move the append to the provisioning path - `internal/runtime/provision.go` has the launch-time context and already owns `briefPath` - and keep `Build` pure. `buildGrok`/`buildPi` can keep their `(LaunchSpec, error)` signature; they just stop writing. Then assert it: a test that reconcile's confirm-launch arm for a grok attempt does not modify the brief file.

**Two things to know rather than act on:**

1. **Rebase needed.** atqamz/hand#411 merged as `9f7f23b` and it touches `briefPrompt` and adds a `Kind` field to `harness.Options`, threaded through the same `buildHarness` seam. Expect conflicts adjacent to your `launchStatement` extraction. The ship/scout authorization sentence #411 adds belongs in `launchStatement`, not in `briefPrompt` - otherwise grok and pi get the report channel but not the delivery grant, which reintroduces this issue for a different string. Please make that so while resolving.

2. **atqamz/hand#448 interacts with this.** That task is adding a brief digest recorded at attempt launch, so `hand promote` can refuse when the supervisor has not rewritten the scout brief. If hand appends to `brief.md` at provision time, the digest has to be taken on a consistent side of the append or hand's own edit reads as the supervisor's revision. I am telling that worker too; you do not need to coordinate, just do not assume the brief file is untouched by hand elsewhere.

PR #454, opened 2026-08-28
fix(registry): make test isolation unforgettable, add prune, refuse managed-tree homes

### LINKED ISSUE #413: hand's own tests register 3366 Fleets in the operator's real registry, and nothing can prune them
## What happens

`hand`'s own test suite registers Fleet homes in the operator's **real** user-local registry at
`/home/atqa/.secondhand/registry.db`. This fleet home's registry now holds 3371 Fleets, of which 3366
are `/tmp/...Test.../002` paths left behind by test runs. There is no way to prune them through the
CLI.

```
$ hand fleet | grep -c '^  f_'
3371
$ hand fleet | grep -c ',/tmp/'
3366
$ hand fleet | grep '^  f_' | grep -v ',/tmp/'
  f_8dbc6c354fc19c2d1179daf3174dacb7,/home/atqa/.no-mistakes/worktrees/.../.tmp-cli/home.qsyQvL,missing,false,...
  f_9005c11de54251598cac839b9949b588,/home/atqa/.treehouse/hand-a7a4d5/3/hand,unreadable,false,...
  f_970d480f94f551a294df5ac7439a2802,/home/atqa/.no-mistakes/worktrees/.../.tmp-cli/home.QEpuKt,missing,false,...
  f_cf12b2a43e88dc7f4fcce131a7aaf00a,/home/atqa/github/atqamz/hand,ready,true,...
  f_f455a789a08334144cbbb2e6996641b0,/home/atqa/.no-mistakes/evidence/.../cli-home,unreadable,false,...
```

Exactly one Fleet is real and `ready`. Sample of what the rest look like - the test names are right
there in the paths:

```
f_000175f6f8c73299dda4577b1df51e9e,/tmp/nix-shell.0zUHVd/TestClassifiedRoutingRefusalsStopBeforeStateAndProvisioningmiss110351637/002,missing
f_001623e06272af7661d76ec1b50d3092,/tmp/nix-shell.Hfq6y0/TestWatchArmProbeInterruptionWhileHerdrIsBlocked1858192080/002,missing
f_0056f881ab4d45bab9fb70290961b78b,/tmp/nix-shell.pZO7AC/TestSendDoesNotCreateRecordBeforeComposerMutation576667179/002,missing
f_00623e4f370cf89b88af14b229c3ff42,/tmp/nix-shell.wJzH5k/TestPRCommand2663354368/005,missing
```

## Why it matters

The bundled `secondhand` skill instructs the supervisor to "use `hand fleet` to inspect user-local
discovery when the current invocation context is unclear", and to stop runtime and mutation commands
if Hand reports a duplicate or identity mismatch. Both instructions are now impractical: the command
emits 3371 rows to answer a question about one home, and any duplicate or identity reasoning has to
run against 3366 dead entries. I hit this while tracing a real cross-fleet worktree problem
(atqamz/hand#403, atqamz/hand#412) and could only use the output by piping it to a file and grepping.

Two of the surviving non-`/tmp` entries are themselves worth noticing, because they show the same
isolation gap outside `/tmp`:

- `f_9005c11de54251598cac839b9949b588` is registered with its home at
  `/home/atqa/.treehouse/hand-a7a4d5/3/hand` - a **Treehouse worktree slot** registered as a Fleet
  home, state `unreadable`. Something ran a home-creating command inside a worker's worktree.
- Two `.no-mistakes` `.tmp-cli/home.*` scratch directories and one `.no-mistakes/evidence/.../cli-home`
  are registered as Fleets.

Separately, a Treehouse lease in `/home/atqa/.treehouse/hand-422b27/5/hand` is held by
`hand:f_167a403f6e12d103a5d310cc10fecedc:340-scout`, and `f_167a403f6e12d103a5d310cc10fecedc` **does
not appear in the registry at all** - an unregistered Fleet identity holding a live lease on worktrees
rooted in this home's own `projects/hand/.git`. See atqamz/hand#412.

## Fix direction

1. **Isolate the registry in tests.** Registration must resolve through an overridable path (env var
   or an explicit parameter threaded from the home), and the test harness must set it to a per-test
   temporary location. Any test that reaches `~/.secondhand/registry.db` should be impossible, not
   merely discouraged - a single missed override cost this home 3366 rows.
2. **Give the CLI a way to prune.** `hand fleet` should be able to drop entries whose home is `missing`
   (the path does not exist), with an explicit flag rather than silently. Without this, an operator has
   no recovery from the state this repository's own tests created.
3. **Refuse to register a home inside a Treehouse worktree or another Fleet's managed tree.** The
   `hand-a7a4d5/3/hand` entry should never have been creatable.
4. Consider making `hand fleet` default to a summary - counts by state, plus the `ready`/current
   entries - with full enumeration behind a flag. 3371 rows is not orientation.

## Acceptance

- Running the full test suite adds zero entries to the user's real registry; a test asserts this.
- An operator can prune `missing` entries with one documented command.
- Registering a Fleet home inside a Treehouse worktree is refused.

Related: atqamz/hand#412

atqamz:
## Summary

- `secondhand.Home()` now refuses (`ErrHomeNotOverridden`) instead of falling back to the operator's real `~/.secondhand` when `SECONDHAND_HOME` is unset in a `-tags=test` binary - the fallback branch that leaked 3366 rows into the real registry is unreachable in a test build, not merely unset by convention. Trace, before any code change, is in this issue's discussion.
- `hand fleet prune` reports (default) or removes (`--apply`) Fleets already classified `missing` - never `ready`, never `unreadable`, never the current home. `registry.MissingFleets`/`registry.Prune` are the underlying API.
- `hand init` now refuses to register a home inside Hand's own Treehouse worktree pool (`worktree.PoolsRoot()`) or inside another Fleet's `projects/` tree, checked structurally against the filesystem rather than the registry, per `docs/adr/the-landed-work-guard-reads-the-work-not-the-record.md`.
- Fix direction 4 from the issue (`hand fleet` summary-by-default) is left out of scope, as a separate UX decision.
- `docs/adr/a-test-build-cannot-resolve-the-real-secondhand-root.md` records the isolation-mechanism decision; `docs/testing-invariants.md` INV-REG-5/6/7 now point at their tests instead of reading "violated today"/"not yet implemented".

Closes atqamz/hand#413

## Test plan

From this worktree, real output:

```
$ make lint
go vet ./...
golangci-lint run
0 issues.
go run ./tools/commentlint .
$ make test
ok  	github.com/atqamz/hand/cmd	149.433s
...
ok  	github.com/atqamz/hand/internal/registry	7.283s
ok  	github.com/atqamz/hand/internal/secondhand	1.043s
ok  	github.com/atqamz/hand/internal/worktree	7.087s
(all packages ok)
$ make e2e
ok  	github.com/atqamz/hand/tests/e2e	68.823s
```

Manual demo in a scratch registry created for this PR (`SECONDHAND_HOME` pointed at a throwaway temp dir, never `~/.secondhand`): seeded a `ready`, an existing-but-unreadable, and a `missing` Fleet, then:

```
$ hand fleet prune
applied: false
candidates[1]{id,home,state,locations}:
  f_e7e937c7897555c2a2b9f8115089d54e,.../missing,missing,.../missing
help[1]:
  - Run `hand fleet prune --apply` to remove exactly these Fleets; nothing was changed

$ hand fleet prune --apply
applied: true
removed[1]{id,home,state,locations}:
  f_e7e937c7897555c2a2b9f8115089d54e,.../missing,missing,.../missing

$ hand fleet
fleets[2]{id,home,state,current,locations}:
  f_...,.../ready,ready,false,.../ready
  f_...,.../unreadable,unreadable,false,.../unreadable
```

The ready and unreadable rows survived untouched; a second `--apply` was a true no-op (`removed[0]`). Also demoed both registration refusals (a Treehouse pool slot and a Fleet's `projects/` tree) returning `exit: 3`, `kind: precondition`, with a `help` line.

The real `~/.secondhand/registry.db` mtime is unchanged across this entire session (checked before and after every step) - nothing in this PR's tests or manual verification ever wrote to it.

### DIFF
--- a/cmd/fleet.go
+++ b/cmd/fleet.go
@@ -61,6 +61,73 @@ func newFleetCmd() *cobra.Command {
 		},
 	}
 	cmd.AddCommand(newFleetHerdrCmd())
+	cmd.AddCommand(newFleetPruneCmd())
+	return cmd
+}
+
+// Opt-in only, and only ever by classification: bare `hand fleet prune` reports the Fleets it would
+// drop and drops nothing; --apply removes exactly those. Never a side effect of any other command,
+// and never anything short of the current classification (see registry.MissingFleets).
+func newFleetPruneCmd() *cobra.Command {
+	var apply bool
+	cmd := &cobra.Command{
+		Use:   "prune",
+		Short: "Report or remove registered Fleets classified missing",
+		Args:  usageArgs(cobra.NoArgs),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			registryPath, err := registry.Path()
+			if err != nil {
+				return err
+			}
+			currentHome, _ := home.Resolve()
+
+			var fleets []registry.Fleet
+			if apply {
+				registryDB, err := registry.OpenAt(registryPath)
+				if err != nil {
+					return err
+				}
+				defer func() { _ = registryDB.Close() }()
+				fleets, err = registryDB.Prune(currentHome)
+				if err != nil {
+					return err
+				}
+			} else {
+				registryDB, err := registry.OpenReadOnlyAt(registryPath)
+				switch {
+				case errors.Is(err, registry.ErrRegistryMissing):
+				case err != nil:
+					return err
+				default:
+					defer func() { _ = registryDB.Close() }()
+					fleets, err = registryDB.MissingFleets(currentHome)
+					if err != nil {
+						return err
+					}
+				}
+			}
+
+			doc := axi.Doc{}
+			doc.Field("registry", registryPath)
+			doc.Bool("applied", apply)
+			label := "candidates"
+			if apply {
+				label = "removed"
+			}
+			rows := make([][]string, 0, len(fleets))
+			for _, fleet := range fleets {
+				rows = append(rows, []string{fleet.ID, fleet.Home, string(fleet.State), strings.Join(fleet.Locations, "\n")})
+			}
+			doc.Rows(label, []string{"id", "home", "state", "locations"}, rows)
+			if apply {
+				doc.Help("Run `hand fleet` to confirm the registry now holds only what these Fleets left behind")
+			} else {
+				doc.Help("Run `hand fleet prune --apply` to remove exactly these Fleets; nothing was changed")
+			}
+			return doc.Render(cmd.OutOrStdout())
+		},
+	}
+	cmd.Flags().BoolVar(&apply, "apply", false, "remove the reported Fleets instead of only reporting them")
 	return cmd
 }
 
--- a/cmd/fleet_test.go
+++ b/cmd/fleet_test.go
@@ -2,6 +2,7 @@ package cmd
 
 import (
 	"bytes"
+	"os"
 	"path/filepath"
 	"strings"
 	"testing"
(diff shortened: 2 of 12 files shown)


atqamz:
Reviewed the substance. This is the strongest delivery of the run and I have no design changes. CI is failing separately - see the note at the end.

Three things this got right that the issue could easily have been under-delivered on:

**Isolation is genuinely unforgettable, not merely defaulted.** `secondhand.Home()` returning `ErrHomeNotOverridden` under `testtag.Present` makes the real-home fallback *unreachable* from a test binary. The issue asked for "impossible, not merely discouraged" and an override a test can forget to set would have been the same bug one layer down. Reusing `internal/testtag` rather than inventing a second discriminant is right.

**The acceptance test is white-box, and it had to be.** The issue asked for "a test asserts the suite adds zero entries to the real registry", and the obvious reading - read the real registry and check - is itself a test that touches the real registry. `TestHomeRefusesWithoutOverrideUnderTestBuild` proves the property at the only place it can be proven without doing the thing it forbids. Saying so in the report rather than quietly picking one was the right call.

**Prune re-classifies under the write lock** instead of trusting the list `MissingFleets` returned to the caller, so only what is still missing at delete time is deleted. And it is `StateMissing` only - never `unreadable`, never `present-but-not-ready`, never the current home.

The registration refusal being structural - `worktree.PoolsRoot()` and a walk for `projects/`, checked against the filesystem rather than the registry - is also correct, precisely because a stale registry is the condition being cleaned up.

**One non-blocking note.** `newFleetPruneCmd` does `currentHome, _ := home.Resolve()`, discarding the error. If resolution failed, `currentHome` is empty and the current home loses its `Current` flag, so the `!fleet.Current` guard stops protecting it. In practice `StateMissing` already makes that unreachable - you cannot be standing in a home whose path does not exist - so nothing is at risk today. But a discarded error in the one command that deletes rows is worth either handling or annotating with why it is safe.

**One consequence worth recording somewhere,** because it changes how workers operate: a Treehouse slot can no longer become a Fleet home, so a worker can no longer `hand init` a scratch home inside its own worktree. That is exactly the refusal the issue asked for - `f_9005c11de54251598cac839b9949b588` is that bug - but it means scratch fleet homes now have to live in a temp directory. Worth a line in the ADR's consequences if it is not already there. I am updating my own brief wording accordingly.

**CI:** `Test (ubuntu-latest)`, `Test (ubuntu-24.04-arm)` and `Nix build (x86_64-linux)` are failing while lint and contract pass. Given what changed, the likely cause is a CI entry point that builds or runs with the test tag without `SECONDHAND_HOME` set - which your refusal correctly converts from a silent real-home fallback into a hard failure. That is the change working, and it means those entry points need the override `make test` already supplies. Sent to the worker.

PR #456, opened 2026-08-28
fix(promote): refuse to launch a ship attempt against an unrevised brief

### LINKED ISSUE #448: fix(promote): a promoted scout relaunches against the scout's own brief, which forbids shipping
## What happens

`hand promote <id>` turns a completed scout into a ship task and immediately launches the ship attempt. The new attempt is launched against the same brief path, `data/<id>/brief.md`, which still holds the **scout** brief.

A scout brief says what a scout brief has to say:

```
This is a scout. The deliverable is a report. Do not change behavior.
...
## Non-goals
- Any behavior change, test change, or refactor. This is investigation only.
...
- No commit, no push, no PR. The report is the deliverable.
```

So the ship worker starts with an instruction sheet that forbids the work it was just promoted to do.

## Observed

```
$ hand promote 410-lock-holder-misattribution
id: 410-lock-holder-misattribution
result: promoted
kind: ship
was: scout
worktree: /home/atqa/.secondhand/pools/hand-5dc40c783afe/.treehouse/hand-a7a4d5/2/hand

$ hand status 410-lock-holder-misattribution --fields id,kind,state,attempt_lifecycle,attempt_ordinal
kind: ship
state: working
attempt_lifecycle: running
attempt_ordinal: 2
```

Attempt 2 was already running before the supervisor could rewrite the brief. Recovered with `hand send`, but only because the supervisor happened to check the attempt state straight after promoting.

## Why it is not just an operator mistake

`hand promote` is the command whose entire purpose is the scout-to-ship transition, and the brief is the one artefact that must change across it. Nothing in the command's output mentions the brief, and nothing pauses before launching. The bundled skill's `references/planning-and-briefs.md` describes what a ship brief needs and says a scout brief must direct its report to `$HAND_HOME/data/<id>/report.md`, so the two briefs are known to be different documents - but the same path holds both, and promotion silently reuses whichever is there.

The failure is quiet in the worst way: the worker is not blocked and reports nothing unusual. It reads a coherent brief and does what it says, which is to investigate and not ship. A supervisor who does not check within the first minute sees a ship task that produced a report and no PR.

## Fix directions, not settled

- Refuse to promote until the brief is not the one attempt 1 ran against - a recorded hash would do - and name the rewrite as the treatment.
- Promote without launching, so the supervisor writes the ship brief and dispatches deliberately. This matches the `scout` and `ship` kinds being separate deliverables rather than phases of one run.
- Keep launching, but pass the ship attempt a distinct path, `data/<id>/ship-brief.md`, and refuse if it is absent.

The first and second both make the missing artefact impossible to miss. The third changes a convention the skill documents, so it is the widest.

## Acceptance

- Promoting a scout cannot silently start a ship attempt against the scout's brief.
- Whatever the mechanism, the supervisor learns it must supply ship instructions before work begins, not after.

atqamz:
## Summary

- Records a digest of the brief at attempt launch (`hand spawn`, `hand reopen`, `hand promote`), alongside `planned_against` on the same `attempt` row.
- `hand promote` refuses, and launches nothing, when the scout attempt's recorded digest still matches `data/<id>/brief.md`'s current digest - naming the file and "rewrite it as a ship brief" as the treatment. An attempt recorded before this change carries no digest and is never treated as unchanged.
- The digest excludes hand's own grok/pi launch-statement appendix (atqamz/hand#418, in review as of this writing), so that append at provision time is never mistaken for a supervisor revision.
- Adds an ADR and a `docs/testing-invariants.md` entry per the repo's convention.

Closes atqamz/hand#448

## Test plan

- [x] `make lint` - clean (`go vet`, `golangci-lint`, `commentlint`)
- [x] `make test` - full suite passes with `-race`
- [x] `make e2e` - passes
- [x] Manual scratch-fleet demonstration via a temporary e2e test (deleted after use):
  - Promote against the untouched scout brief:
    ```
    $ hand promote task-1
    exit 3
    error: brief at data/task-1/brief.md is unchanged since the scout attempt launched; rewrite it as a ship brief before promoting
    ```
  - Promote after rewriting the brief:
    ```
    $ hand promote task-1
    exit 0
    result: promoted
    kind: ship
    was: scout
    ```

### DIFF
--- a/cmd/promote_test.go
+++ b/cmd/promote_test.go
@@ -8,6 +8,7 @@ import (
 	"testing"
 	"time"
 
+	"github.com/atqamz/hand/internal/brief"
 	"github.com/atqamz/hand/internal/faketool"
 	"github.com/atqamz/hand/internal/harness"
 	"github.com/atqamz/hand/internal/herdr"
@@ -29,7 +30,17 @@ func promoteHerdr(agent, status string) faketool.Herdr {
 	}
 }
 
+const promoteScoutBriefContent = "implement the fix"
+
 func setupPromoteHome(t *testing.T, oldWorktree, newWorktree string, herdr faketool.Herdr) string {
+	t.Helper()
+	return setupPromoteHomeWithScoutBriefDigest(t, oldWorktree, newWorktree, herdr, "")
+}
+
+// scoutBriefDigest lets a caller simulate a scout attempt launched under atqamz/hand#448's digest
+// tracking, without going through UpdateAttempt: BriefDigest is write-once execution identity, so
+// it can only be set at the attempt's creation, same as PlannedAgainst.
+func setupPromoteHomeWithScoutBriefDigest(t *testing.T, oldWorktree, newWorktree string, herdr faketool.Herdr, scoutBriefDigest string) string {
 	t.Helper()
 	useFastLaunchPolling(t)
 	t.Setenv("HAND_HARNESS", harness.Claude)
@@ -41,7 +52,7 @@ func setupPromoteHome(t *testing.T, oldWorktree, newWorktree string, herdr faket
 	if err := os.WriteFile(filepath.Join(home, "data", "task-1", "report.md"), []byte("scout findings"), 0o644); err != nil {
 		t.Fatal(err)
 	}
-	if err := os.WriteFile(filepath.Join(home, "data", "task-1", "brief.md"), []byte("implement the fix"), 0o644); err != nil {
+	if err := os.WriteFile(filepath.Join(home, "data", "task-1", "brief.md"), []byte(promoteScoutBriefContent), 0o644); err != nil {
 		t.Fatal(err)
 	}
 	if err := project.Add(home, project.Project{Name: "myproj", URL: "https://example.com/myproj.git", Mode: project.ModeDirectPR}); err != nil {
@@ -51,7 +62,7 @@ func setupPromoteHome(t *testing.T, oldWorktree, newWorktree string, herdr faket
 		t.Fatal(err)
 	}
 
-	if err := writeTaskAttempt(t, home, state.Task{ID: "task-1", Project: "myproj", Kind: state.KindScout}, state.Attempt{Lifecycle: state.AttemptRunning, Worktree: oldWorktree, Herdr: state.Herdr{WorkspaceID: "wA", TabID: "wA:tOld", PaneID: "wA:pOld"}}); err != nil {
+	if err := writeTaskAttempt(t, home, state.Task{ID: "task-1", Project: "myproj", Kind: state.KindScout}, state.Attempt{Lifecycle: state.AttemptRunning, Worktree: oldWorktree, Herdr: state.Herdr{WorkspaceID: "wA", TabID: "wA:tOld", PaneID: "wA:pOld"}, BriefDigest: scoutBriefDigest}); err != nil {
 		t.Fatal(err)
 	}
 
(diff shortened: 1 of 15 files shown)


atqamz:
Reviewed. The refusal is correctly placed - before the lock and before `PromoteTask`, so a refused promote launches nothing, which was the invariant most at risk here. The empty-digest branch is handled exactly right and for the stated reason: a scout attempt predating digest recording is not an unchanged brief, and refusing every pre-existing task would have been a worse failure than the one being fixed. And using `Precondition` is correct because, unlike atqamz/hand#422's case, this one genuinely is satisfiable - rewrite the brief and the same command runs.

Picking up the atqamz/hand#418 interaction by stripping hand's own appendix before hashing, with a test proving the refusal still fires when only the appendix was added, is the right resolution and more than I asked for.

**One change: the marker must have one definition, not two.**

This PR adds:

```go
// AppendMarker ... It must stay byte-identical to internal/harness's own copy
const AppendMarker = "hand appended the block below at launch time; it is not part of the supervisor's brief above."
```

and PR #453 independently adds:

```go
const briefAppendMarker = "hand appended the block below at launch time; it is not part of the supervisor's brief above."
```

They are byte-identical today. A comment asking two constants to stay in sync is not a mechanism, and the failure mode if they drift is silent: `Digest` stops finding the boundary, hand's own append reads as the supervisor's revision, and `promote` launches exactly when it should refuse. That is this issue's bug, reintroduced by the fix for it.

`internal/brief` is the right home - `internal/harness` already imports it for `brief.ExecutionClass`. Keep `brief.AppendMarker` exported here as the single source of truth, and have the append side consume it rather than declaring its own.

Since both PRs are open, whichever merges second does the wiring. I am telling PR #453's worker to import `brief.AppendMarker` and delete `briefAppendMarker`, and to keep that in mind because it is also being asked to move `appendLaunchStatement` into `internal/runtime/provision.go`. If #453 lands first, rebase and delete nothing; if this lands first, #453 does the deletion.

Add a test that pins the coupling rather than relying on the comment: assert that a brief carrying a real appendix - produced by the append path, not a hand-written literal - digests identically to the same brief without one. The existing `TestDigestUnaffectedByHandsOwnAppendix` builds the appendix from a string literal, so it would still pass if the two constants diverged.

atqamz:
Pushed the marker-comment fix: `brief.AppendMarker`'s doc comment now says it is the one definition the append side imports, not two copies kept in sync by convention.

On the coupling test - I can't write it yet. `appendLaunchStatement` (the real append path) doesn't exist in this branch: it lives only in #453, which is unmerged, and per your note it's also moving to `internal/runtime/provision.go` there rather than staying in `internal/harness`. Anything I could call today would either not compile against this branch alone, or (if I duplicated the function here to make it compile) would double-implement work #453 is already doing and conflict at merge - worse than the gap it's meant to close.

Deciding this myself since it's reversible: I'll add the real-append-path test as a follow-up commit to this same PR (still unmerged) once #453 lands and I rebase onto it, calling whatever `internal/runtime/provision.go` exposes rather than a hand-written literal. Until then `TestDigestUnaffectedByHandsOwnAppendix` stays as the within-package regression check - it can't catch a cross-package format drift, which is exactly the gap this follow-up closes.

atqamz:
Rebased onto main (now includes #453/d5968f1). Deleted \`harness.briefAppendMarker\`; \`appendLaunchStatement\` now builds the appendix from \`brief.AppendMarker\`, its one remaining definition.

Added the coupling test in \`internal/harness/harness_test.go\`: \`TestAppendPromptToBriefLeavesBriefDigestUnchanged\` calls the real \`AppendPromptToBrief\` for grok and pi, then asserts \`brief.Digest\` is unchanged before/after - no hand-written literal standing in for the append. Verified it actually catches drift by perturbing the appendix delimiter locally and confirming the test fails, then reverted.

\`make lint\`, \`make test\` (race), and \`make e2e\` all pass on the rebased branch. Pushed as 9723ed9 (force-with-lease, since this is a rebase). Not merged.

atqamz:
Rebase and marker unification reviewed. Accepted with no further changes.

`harness.briefAppendMarker` is gone, `appendLaunchStatement` builds from `brief.AppendMarker`, and `TestAppendPromptToBriefLeavesBriefDigestUnchanged` closes the loop the way I wanted: it calls the real `AppendPromptToBrief` for both grok and pi and asserts `brief.Digest` is unchanged across it, so the two sides are now coupled by a test rather than by a comment. Verifying it by perturbing the delimiter and watching it fail before restoring is the right way to prove a regression test, and that is twice this fleet has done it unprompted today.

Merging once CI is green.

atqamz:
Fixed: both refusal-message assertions in \`cmd/promote_test.go\` hardcoded \`"data/task-1/brief.md"\` with a forward slash; on windows-latest the error renders it with a backslash, so the string-contains check failed even though the refusal fired correctly. Both now build the expected substring with \`filepath.Join("data", "task-1", "brief.md")\`.

Checked every other new assertion in this PR's diff for the same hardcoded-separator pattern (\`cmd/promote_test.go\`, \`cmd/spawn_test.go\`, \`internal/brief/brief_test.go\`, \`internal/harness/harness_test.go\`) - the remaining slash-containing string literals are all inert placeholder text baked into fabricated brief/appendix content (e.g. \`/tmp/task-1.status\`), never compared against a real OS-rendered path, so none of them are exposed to this.

\`make lint\`, \`make test\` (race), \`make e2e\` all pass. Pushed as 400268d.

Issue #423, opened 2026-08-26
hand pr cannot record a PR delivered to a sibling repository, and the only offered escape misrepresents the project topology


atqamz:
## What happens

```
$ hand pr butler-broker-allowlist-rujak-release https://github.com/yes2games/butler/pull/92
error: "PR https://github.com/yes2games/butler/pull/92 belongs to yes2games/butler, not project
        yes2infra's repo (yes2games/yes2infra) and no upstream is declared for it"
kind: precondition
exit: 3
```

The refusal is correct as a guard - a PR from an unrelated repo should not be recordable - but it has no correct escape for a legitimate case, so the task can never record where its work actually landed.

## The legitimate case

Task `butler-broker-allowlist-rujak-release` was dispatched against project `yes2infra` to add a broker allowlist entry at `pulumi/worker/config/broker.json`. Between the brief being written and the worker starting, yes2games/yes2infra#435 **deleted that file** and moved the Worker to `yes2games/butler`. The file, the `BrokerEntry` shape, the existing entry and the vitest gate all moved verbatim; ADR-0015 carries an update saying so; and nothing in `yes2infra` needed changing, since it still owns the `/token` route and the Access bypass.

The worker made the right call: it followed the code, delivered yes2games/butler#92 (squash `c1a88f3`), documented the deviation as reversible, and verified it thoroughly. The outcome is correct and the state record is wrong - `pr: unknown` forever.

## Why the offered escape does not fit

The error names one: "no upstream is declared for it". But `hand project upstream` is documented as "Declare the upstream repo a fork project opens its PRs against". `yes2infra` is not a fork of `butler`; they are sibling repositories in the same organization. Declaring butler as yes2infra's upstream would put a false topology in the registry and change PR resolution for **every** future yes2infra task, to record one PR on one task. That is a worse outcome than the missing record.

## Ask

Let a task record a PR in a different repository than its project's, when the supervisor asserts it deliberately. The safety property worth keeping is "not by accident", not "not at all" - so an explicit opt-in flag on `hand pr` (`--cross-repo`, or requiring the full `owner/repo#N` form rather than bare acceptance) preserves the guard while making the true outcome recordable.

Consider also carrying the reason: a cross-repo delivery is always a scope deviation worth reading later, and today the only place it survives is the worker's free-text report.

## Scope

[...]

PR #457, opened 2026-08-28
fix(runtime): close the worktree release gate on evidence

### LINKED ISSUE #424: Terminal-task resource release is still a closed circle: teardown demands a recorded PR, hand pr refuses on a torn-down task, abandon-worktree correctly no-ops
atqamz/hand#407 asked for two things. Its first - distinguishing proven PR absence from unknown observation - landed in PR atqamz/hand#415 and is verified. Its second was "one command that releases a terminal task's still-recorded resources without inventing an attempt", and that one did **not** land. I checked it against the exact task that motivated the issue, on `main` at `45fab52743` with #415 in the tree, and the circle still closes. It is narrower now, and it closes through a different door.

## Verified repro, all three doors, on task `402-report-pipeline-fix`

**Door 1 - `hand teardown`:**

```
$ hand teardown 402-report-pipeline-fix
error: "refusing to return terminal task \"402-report-pipeline-fix\" worktree
 /home/atqa/.treehouse/hand-a7a4d5/12/hand: commit safety is not proven: whether worktree ...
 still holds the only copy of any commit could not be observed: the branch records upstream
 origin/402-report-pipeline-fix and no remote-tracking ref for it survives, so what the remote
 holds is no longer recorded here, and no pull request is recorded for this task, so no pushed
 head can hold them either; observed by running \"git rev-list --count HEAD --not --remotes\" ...
 the return is withheld because the question went unanswered, not because work was found at risk"
kind: precondition
```

This gate is right and should not be weakened. The message is a model of a good refusal: it names the observation it ran, distinguishes "unanswered" from "at risk", and names the missing evidence - a recorded PR.

**Door 2 - supply the evidence it asks for:**

```
$ hand pr 402-report-pipeline-fix https://github.com/atqamz/hand/pull/405
error: task 402-report-pipeline-fix is torn down; run hand reopen 402-report-pipeline-fix before recording a PR on it
kind: precondition
```

**Door 3 - attest and relinquish:**

```
$ hand reconcile 402-report-pipeline-fix --abandon-worktree
result: healthy
action: none
```

A no-op, and correctly so: the flag refuses any lease an observation can still prove, and this lease is observable. `treehouse status` shows slot 12 `leased` with five live processes.

`hand reopen` is the only remaining move, and it is what the issue explicitly ruled out - it invents a new attempt to record a fact about an old one. In this fleet home it is also unavailable, since `hand spawn`/`hand reopen` are frozen for project `hand` by atqamz/hand#412.

## The shape of the deadlock

A partial teardown moved this task to **torn down** - which is the state `hand pr` refuses on - while leaving its Treehouse lease and Herdr pane still recorded, which is the state only a teardown can release. Teardown then demands evidence that only `hand pr` can supply. Each command's precondition is individually defensible; together they are unsatisfiable.

## The evidence exists, and no command can accept it

I verified by hand, read-only, that nothing is at risk. The worktree holds one unpushed commit, `6fdd6c7 fix reconciliation for missing worker reports`, touching `cmd/reconcile.go`, `internal/runtime/liveness_test.go` (+49) and `internal/runtime/reconcile.go` - the same three files PR atqamz/hand#405 landed. Both test functions it adds are present on `origin/main`:

```
$ git grep -c "func TestReconcileConvergesDonePaneWithoutWorkerReport(" origin/main -- internal/runtime/liveness_test.go
origin/main:internal/runtime/liveness_test.go:1
$ git grep -c "func TestReconcileKeepsDonePaneWithTerminalWorkerReport(" origin/main -- internal/runtime/liveness_test.go
origin/main:internal/runtime/liveness_test.go:1
```

The commit's own patch-id does not match anything upstream because #405 was squash-merged, which is exactly why `git rev-list --count HEAD --not --remotes` cannot answer and the gate withholds.

## Ask

Let the operator supply the evidence the gate already names, without resuming work. Either of these closes it:

1. **Allow `hand pr` on a torn-down task.** Recording where work landed is not resuming it, and the command is already write-once and validated. This is the smaller change and it makes the existing refusal text actionable.
2. **Let teardown accept an explicit attestation** for the commit-safety question specifically - the same shape `--abandon-worktree` already uses for the ownership question, and with the same discipline: refuse it whenever the question *can* be answered, so it can only ever be used where hand is genuinely blind.

Do not relax the commit-safety gate itself. The defect is that its stated remedy is unreachable, not that it asks too much.

Related: atqamz/hand#407 (closed; first ask landed), atqamz/hand#412 (why `hand reopen` is not available here), atqamz/hand#422 and atqamz/hand#423 (same family - `hand` state unable to record a true outcome).

### LINKED ISSUE #428: teardown cannot release a worktree that provably has no PR: a detached pool HEAD with no recorded branch reports unknown instead of absent
## Context

Distinct from atqamz/hand#424, which is the *terminal-task* circle where the lease is observable.
This one is an *active* attempt whose worktree sits on the pool's detached HEAD.

A pool slot starts detached; a branch only exists once the worker creates one. A worker that
legitimately never created a branch - it paused, or it was stopped before committing - leaves an
attempt with `Branch: ""` and a detached worktree. `hand teardown` then refuses, permanently:

```
error: "whether 23-launcher-activation has a pull request could not be observed, so its worktree
is not released: resolve the branch to search for in
/home/atqa/.treehouse/omanixy-4a68e2/3/omanixy: git HEAD is detached or unreadable:
git symbolic-ref --short -q HEAD: : run managed command git: exit status 1"
```

`internal/runtime/pr.go` is right that a detached worktree "cannot answer at all - not evidence
that no branch, and so no PR, ever existed". But it is answerable here, and hand already has every
observation it needs to answer it:

- HEAD is detached **and** `active.Branch` is empty, so no branch was ever recorded for this
  attempt, and
- `git rev-list --count HEAD --not --remotes` is `0`, which `internal/worktree/worktree.go`
  already computes as `localOnlyCommitCount`, so this worktree holds no commit that is not already
  on a remote.

Together those disprove a pull request rather than leaving it unknown: there is no branch to have
opened one from and no commit that could only have reached GitHub through one. Reporting `unknown`
here is a false negative, and it is the only thing standing between a provably empty worktree and
its release.

## The escape hatch works, but only in one undocumented order

Three tasks in this state (`23-launcher-activation`, `24-font-wallpaper-ownership`,
`39-tray-sni-proof`) were recovered, so this is not a hard lock - but the sequence is not
discoverable from any error message:

1. `hand teardown <id>` - fails at PR observation as above. Nothing suggests `--force`, and
   `--force`'s own help ("skip landed-work checks") does not read as covering "cannot resolve a
   branch".
2. `hand reconcile <id> --abandon-worktree` - refuses, correctly, with *"end this attempt with
   `hand teardown <id>` if it is still active"*. Which is step 1, which just failed. Read on its
   own this is a closed circle.
3. `hand teardown <id> --force` - gets past PR observation and fails at the *next* step,
   `prove worktree ownership before return`, because the lease lives in a pre-#404 pool hand no
   longer looks at. Crucially, this failure is what **records the teardown decision** that step 2
   was waiting for.
4. `hand reconcile <id> --abandon-worktree` - now succeeds.

So `--abandon-worktree`'s precondition is not "teardown succeeded" but "teardown got far enough to
record a decision and then failed at the worktree step" - which its help text does say, in
*"the shape a teardown interrupted at the worktree step leaves behind"*. It just cannot be reached
from step 2's message, which names only step 1.

## Asks

1. **Let PR observation disprove.** When HEAD is detached, no branch is recorded, and
   `localOnlyCommitCount` is 0, report the PR as observed-absent instead of unknown. Then teardown
   releases these worktrees with no flags at all, which is the actual fix.
2. **Make step 2 point at step 3.** When `--abandon-worktree` refuses an active attempt, say that a
   `hand teardown` that fails *before* the worktree step does not record the decision, and name
   `--force` for the case where the blocking check is one the operator has independently verified.
3. **Say what `--force` skips.** "skip landed-work checks" does not tell an operator that branch
   resolution and PR observation are among them.

Worth stating what forcing actually risked here, since that is the judgement the flag exists to
protect: all three worktrees were verified clean (`git status --porcelain` empty), at zero
local-only commits, and byte-identical to snapshots taken before, so `--force` skipped a check
whose answer was already established three other ways. That verification is exactly what ask 1
would let hand do for itself.

Related: atqamz/hand#424, atqamz/hand#412, atqamz/hand#411.

atqamz:
## Summary

- `internal/runtime/pr.go` now proves a pull request absent, rather than answering `unknown`, on a detached worktree with no recorded branch and no commit missing from a remote-tracking ref - the exact evidence `worktree.ObserveCommitSafety` already computes. `checkLandedWork` (`teardown.go`) treats the same three-condition proof as no committable work to land, so `hand teardown` releases such a worktree honestly instead of demanding an unanswerable PR search.
- `hand pr` (`RecordPR`) no longer refuses a terminal task. Recording where work landed is not resuming it: the write stays write-once, still validates the URL against the project and GitHub, and touches no lifecycle column. `hand pr`'s help text no longer suggests `hand merge` on a terminal task, since there is no active attempt left for it to act on.
- Adds an ADR documenting both fixes as the same worktree-release gate approached from different callers, and explicitly rejecting a teardown attestation flag (atqamz/hand#424's second ask) in favor of proving absence from observations hand already holds.

Closes atqamz/hand#424
Closes atqamz/hand#428

## Test plan

```
$ make lint
go vet ./...
golangci-lint run
0 issues.
go run ./tools/commentlint .

$ make test
ok      github.com/atqamz/hand/cmd                       138.509s
ok      github.com/atqamz/hand/internal/age              (cached)
ok      github.com/atqamz/hand/internal/agentsmd         (cached)
ok      github.com/atqamz/hand/internal/attention        (cached)
ok      github.com/atqamz/hand/internal/axi              (cached)
ok      github.com/atqamz/hand/internal/brief            (cached)
ok      github.com/atqamz/hand/internal/completion       (cached)
ok      github.com/atqamz/hand/internal/faketool         3.967s
ok      github.com/atqamz/hand/internal/filelock         (cached)
ok      github.com/atqamz/hand/internal/ghutil           3.876s
ok      github.com/atqamz/hand/internal/git              4.538s
ok      github.com/atqamz/hand/internal/harness          (cached)
ok      github.com/atqamz/hand/internal/herdr            5.505s
ok      github.com/atqamz/hand/internal/home              (cached)
ok      github.com/atqamz/hand/internal/integration      1.949s
ok      github.com/atqamz/hand/internal/launch           (cached)
ok      github.com/atqamz/hand/internal/notify           (cached)
ok      github.com/atqamz/hand/internal/orientation      (cached)
ok      github.com/atqamz/hand/internal/project          7.640s
ok      github.com/atqamz/hand/internal/registry         4.560s
ok      github.com/atqamz/hand/internal/routing          (cached)
ok      github.com/atqamz/hand/internal/runtime          97.921s
ok      github.com/atqamz/hand/internal/secondhand       1.025s
ok      github.com/atqamz/hand/internal/selfupdate       3.195s
ok      github.com/atqamz/hand/internal/sessionhook      (cached)
ok      github.com/atqamz/hand/internal/shellquote       (cached)
ok      github.com/atqamz/hand/internal/skill            (cached)
ok      github.com/atqamz/hand/internal/state            (cached)
ok      github.com/atqamz/hand/internal/steering         5.656s
ok      github.com/atqamz/hand/internal/store            (cached)
ok      github.com/atqamz/hand/internal/supervision      (cached)
ok      github.com/atqamz/hand/internal/toolchain        (cached)
ok      github.com/atqamz/hand/internal/watcher          45.726s
ok      github.com/atqamz/hand/internal/workerobs        (cached)
ok      github.com/atqamz/hand/internal/worktree         5.195s
ok      github.com/atqamz/hand/tests/edgepublish         4.459s
ok      github.com/atqamz/hand/tests/release             (cached)
ok      github.com/atqamz/hand/tools/commentlint         (cached)

$ go test -tags=e2e,test -count=1 -timeout=10m ./tests/e2e/...
ok      github.com/atqamz/hand/tests/e2e    79.731s

$ make contract
go test -tags=contract -count=1 -timeout=10m ./tests/contract/...
ok      github.com/atqamz/hand/tests/contract    0.464s
```

A scratch-fleet-home reproduction of both issues' exact repros (with real TOON output) will follow as a comment on this PR.

### DIFF
--- a/cmd/pr.go
+++ b/cmd/pr.go
@@ -50,7 +50,11 @@ func newPRCmd() *cobra.Command {
 			doc.Field("id", task.ID)
 			doc.Field("result", result)
 			doc.Field("pr", url)
-			doc.Help("Run `hand merge " + task.ID + "` once this PR's checks are green")
+			// A torn-down task has no active attempt for hand merge to act on (atqamz/hand#424): naming
+			// it here would suggest the task is live again, which recording a PR on it must never do.
+			if task.Lifecycle != state.TaskTerminal {
+				doc.Help("Run `hand merge " + task.ID + "` once this PR's checks are green")
+			}
 			return doc.Render(cmd.OutOrStdout())
 		},
 	}
--- a/cmd/prdetect.go
+++ b/cmd/prdetect.go
@@ -15,9 +15,9 @@ import (
 // attention-is-one-derivation-over-three-channels.md, invariant 1). Empty where no lookup applies.
 func detectPRForStatus(ctx context.Context, home string, history state.TaskHistory) ghutil.PRObservation {
 	t := history.Task
-	// A scout task never answers for a PR - its deliverable is data/<id>/report.md - and a torn-down task's
-	// completion record is already written, so both skip the lookup rather than pay a forge round trip for a
-	// PR recordPR would refuse. The scout half is the short-circuit checkLandedWork opens with.
+	// A scout task never answers for a PR - its deliverable is data/<id>/report.md - and a torn-down
+	// task's completion record is already written, so both skip the lookup rather than pay a forge
+	// round trip a status render has no use for. hand pr can still record one directly (atqamz/hand#424).
 	if t.PR != "" || t.Kind == state.KindScout || t.Lifecycle == state.TaskTerminal {
 		return ghutil.PRObservation{}
 	}
--- a/cmd/task_attempt_test.go
+++ b/cmd/task_attempt_test.go
@@ -3,11 +3,13 @@ package cmd
 import (
 	"encoding/json"
 	"io"
+	"path/filepath"
 	"strings"
 	"testing"
 
+	"github.com/atqamz/hand/internal/faketool"
+	"github.com/atqamz/hand/internal/project"
 	"github.com/atqamz/hand/internal/state"
-	"github.com/spf13/cobra"
 )
 
 func TestStatusInspectsTerminalTaskAndAttemptHistory(t *testing.T) {
@@ -75,37 +77,87 @@ func terminalTaskHome(t *testing.T) string {
 	return home
 }
 
-// Teardown writes the permanent completion record from the state the task had then, so a
-// delivery or a PR recorded afterwards would sit on the row unread. Both name hand reopen.
-func TestDeliverAndPRRefuseATerminalTask(t *testing.T) {
-	for _, tc := range []struct {
-		name string
-		cmd  func() *cobra.Command
-		args []string
-	}{
-		{"deliver", newDeliverCmd, []string{"task-1", "--reason", "handed to upstream"}},
-		{"pr", newPRCmd, []string{"task-1", "https://github.com/o/nsr/pull/7"}},
-	} {
-		t.Run(tc.name, func(t *testing.T) {
-			home := terminalTaskHome(t)
-			cmd := tc.cmd()
-			cmd.SetOut(io.Discard)
-			cmd.SetErr(io.Discard)
-			cmd.SetArgs(tc.args)
-			err := cmd.Execute()
-			if err == nil {
-				t.Fatalf("hand %s accepted a terminal task", tc.name)
-			}
-			if !strings.Contains(err.Error(), "hand reopen task-1") {
-				t.Fatalf("error = %v, want it to name the remedy", err)
-			}
-			got, err := state.Read(home, "task-1")
-			if err != nil {
-				t.Fatal(err)
-			}
-			if got.DeliveredAt != "" || got.PR != "" {
-				t.Fatalf("terminal task was written to: %+v", got)
-			}
-		})
+// Teardown writes the permanent completion record from the state the task had then, so a delivery
+// recorded afterwards would sit on the row unread. hand pr is the one exception (atqamz/hand#424):
+// see TestPRRecordsOnATerminalTaskWithoutReopening.
+func TestDeliverRefusesATerminalTask(t *testing.T) {
+	home := terminalTaskHome(t)
+	cmd := newDeliverCmd()
+	cmd.SetOut(io.Discard)
+	cmd.SetErr(io.Discard)
+	cmd.SetArgs([]string{"task-1", "--reason", "handed to upstream"})
+	err := cmd.Execute()
+	if err == nil {
+		t.Fatal("hand deliver accepted a terminal task")
+	}
+	if !strings.Contains(err.Error(), "hand reopen task-1") {
+		t.Fatalf("error = %v, want it to name the remedy", err)
+	}
+	got, err := state.Read(home, "task-1")
+	if err != nil {
+		t.Fatal(err)
+	}
+	if got.DeliveredAt != "" {
+		t.Fatalf("terminal task was written to: %+v", got)
+	}
+}
+
+// atqamz/hand#424 ask 1: recording where a torn-down task's work landed is not resuming it, so hand
+// pr is the one write allowed past teardown's completion record. Lifecycle must stay terminal, and
+// taskFlags/needsAttention (the fleet's one attention definition) must never key off task.PR.
+func TestPRRecordsOnATerminalTaskWithoutReopening(t *testing.T) {
+	home := terminalTaskHome(t)
+	registerTerminalTaskProject(t, home)
+	const pr = "https://github.com/owner/repo/pull/7"
+	faketool.GH{PRs: []faketool.GHPR{{Number: 7, URL: pr, Repo: "owner/repo", State: "OPEN"}}}.Install(t, faketool.Bin(t))
+
+	cmd := newPRCmd()
+	var out strings.Builder
+	cmd.SetOut(&out)
+	cmd.SetErr(io.Discard)
+	cmd.SetArgs([]string{"task-1", pr})
+	if err := cmd.Execute(); err != nil {
+		t.Fatalf("hand pr refused a terminal task: %v", err)
+	}
+	if strings.Contains(out.String(), "hand merge") {
+		t.Fatalf("output = %q, want no hand merge suggestion: the task has no active attempt for it to act on", out.String())
+	}
+
+	got, err := state.Read(home, "task-1")
+	if err != nil {
+		t.Fatal(err)
+	}
+	if got.PR != pr {
+		t.Fatalf("task.PR = %q, want %q recorded", got.PR, pr)
+	}
+	if got.Lifecycle != state.TaskTerminal {
+		t.Fatalf("task.Lifecycle = %q, want it to stay terminal - hand pr must not reopen the task", got.Lifecycle)
+	}
+
+	// Write-once still holds: a second, different URL is refused exactly like an open task's would be.
+	again := newPRCmd()
+	again.SetOut(io.Discard)
+	again.SetErr(io.Discard)
+	again.SetArgs([]string{"task-1", "https://github.com/owner/repo/pull/8"})
+	if err := again.Execute(); err == nil {
+		t.Fatal("hand pr accepted a second, different PR on a task that already has one recorded")
+	}
+
+	view := taskView{task: got}
+	if needsAttention(view) {
+		t.Fatalf("needsAttention(%+v) = true, want a terminal task with a recorded PR to gain no liveness", view)
+	}
+	if flags := taskFlags(view); len(flags) != 0 {
+		t.Fatalf("taskFlags(%+v) = %v, want none: a recorded-but-unmerged PR renders no flag", view, flags)
+	}
+}
+
+func registerTerminalTaskProject(t *testing.T, home string) {
+	t.Helper()
+	clonePath := filepath.Join(home, "projects", "demo")
+	initGitRepo(t, clonePath)
+	runGitIn(t, clonePath, "remote", "add", "origin", "https://github.com/owner/repo.git")
+	if err := project.Add(home, project.Project{Name: "demo", URL: "https://github.com/owner/repo.git", Mode: project.ModeDirectPR}); err != nil {
+		t.Fatal(err)
 	}
 }
--- a/docs/adr/terminal-release-closes-on-evidence-not-a-new-escape-hatch.md
+++ b/docs/adr/terminal-release-closes-on-evidence-not-a-new-escape-hatch.md
@@ -1,0 +1,108 @@
+# Terminal release closes on evidence, not a new escape hatch
+
+- Date: 2026-08-28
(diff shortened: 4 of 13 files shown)


atqamz:

Both halves land the way the issues asked, and three things stand out.

**The absence proof is conjunctive and fails closed.** `observeDetachedHeadAbsence` requires all three of detached HEAD, empty `active.Branch`, and zero local-only commits, and returns `proven: false` on any inconclusive read - including a read that errors. Splitting `git.IsDetachedHead` out as its own primitive rather than inferring detachment from `CurrentBranch`'s error is the right call for the stated reason: `CurrentBranch` collapses "detached" and "unreadable" into one message, and this proof needs to know which. That is the same unknown-versus-absent distinction atqamz/hand#428 exists to protect, applied to its own inputs.

`TestObservePRStaysUnknownWhenABranchIsRecordedDespiteDetachedHeadWithNoLocalOnlyCommits` and `TestObservePRStaysUnknownWhenDetachedHeadCarriesALocalOnlyCommit` pin the boundary from both sides, and `TestTeardownStillRefusesAShipTaskWhosePRWasNeverOpened` stays untouched. The gate is not weakened anywhere.

**`hand pr` on a terminal task is proven not to grant liveness.** I asked for this to be checked and stated; `TestPRRecordsOnATerminalTaskWithoutReopening` does better than state it - it asserts the lifecycle stays terminal, that write-once still refuses a second different URL, and that `needsAttention` is false for the resulting view. That last assertion is the one that matters and it would have been easy to omit.

**The `TeardownDisposition` detail was not in the brief and is the best catch here.** Recording the no-committable-work outcome as a disposition rather than a boolean, because a retried teardown can reach `completionFor` without calling `checkLandedWork` again and would otherwise fall through to `default` and record "branch merged" for an attempt that never made a branch. That is a real bug avoided in a path neither issue describes.

The ADR's rejected-alternatives section is also right to record *why* the attestation was refused rather than just that it was: the blind spot the attestation would have covered is smaller after this change, so adding it anyway would have made it the path of least resistance for a question hand can now often answer itself.

One note for later, not for this PR: `hand pr`'s reach now extends to terminal tasks, and atqamz/hand#423 wants it to reach a sibling repository. Whoever takes #423 should build on this rather than around it - the validation this PR left intact is exactly what #423 needs to keep while adding a deliberate opt-in.

atqamz:
## Scratch-fleet-home repro (real `git`, `gh`, `treehouse`; real CLI binary built from this branch)

Built in a plain temp dir, never inside a Treehouse worktree (PR #454). Fleet home via `hand init`,
a real clone of `atqamz/hand` registered as a project, two real Treehouse-leased worktrees acquired
directly against the project's own pool (`--root` keyed by a digest of the clone path, isolated from
every other fleet's pool). Task/attempt state was seeded directly with `internal/state` rather than
via `hand spawn`, since spawning a live coding agent for a throwaway demo isn't appropriate; every
command run against that state below is the real `hand` CLI.

### atqamz/hand#424 - the closed circle, reproduced and closed

Fixture: `task-424`, terminal, with a real Treehouse lease and a Herdr pane still recorded, a real
branch commit that is local-only relative to its (simulated-pruned) upstream, and no PR.

**Door 1 - refuses, naming the missing evidence (byte-for-byte the issue's own refusal shape):**

```
$ hand teardown task-424
error: "refusing to return terminal task \"task-424\" worktree .../hand-a7a4d5/1/hand: commit safety
 is not proven: whether worktree .../hand-a7a4d5/1/hand still holds the only copy of any commit could
 not be observed: the branch records upstream origin/424-scratch-repro and no remote-tracking ref for
 it survives, so what the remote holds is no longer recorded here, and no pull request is recorded for
 this task, so no pushed head can hold them either; observed by running \"git rev-list --count HEAD
 --not --remotes\" ...; the return is withheld because the question went unanswered, not because work
 was found at risk"
kind: precondition
exit: 3
```

**Door 2 - `hand pr` now succeeds on the terminal task (this is the fix; before it, this refused with
"run hand reopen task-424 first"):**

```
$ hand pr task-424 https://github.com/atqamz/hand/pull/405
id: task-424
result: recorded
pr: "https://github.com/atqamz/hand/pull/405"
exit=0
```

No `hand reopen` anywhere in this sequence.

**Retrying teardown afterward still refuses** - but now because the gate cross-checked the recorded
PR's actual head commit against the worktree's, and #405 (a real, already-merged PR, deliberately not
a PR opened from this scratch commit, to avoid pushing throwaway content to the public repo) does not
match:

```
$ hand teardown task-424
error: "... and pull request https://github.com/atqamz/hand/pull/405 records head commit 6fdd6c72273c
 instead, so GitHub was never observed holding 5fd20fef52aa; ..."
kind: precondition
exit: 3
```

This is the gate working as designed, not a regression - it is still refusing anything it cannot
prove. The fully-matching case (a PR whose recorded head commit *is* the worktree's HEAD) is what
`TestTeardownCircleClosesOnceHandPRRecordsTheEvidenceItAsksFor` proves end-to-end with a fixture that
can construct that match without a real push.

**No liveness leak** - `hand status` on the fleet after `hand pr` recorded a PR on the terminal task:

```
$ hand status
count: 2
attention: 0
tasks[2]{id,state,reported,age,flags}:
  task-424,unknown,none,unknown,none
  task-428,unknown,none,unknown,none
```

`flags: none` for task-424 despite the recorded PR; `attention: 0`.

### atqamz/hand#428 - detached HEAD, no branch, releases with no flags

Fixture: `task-428`, open/running, worktree on the pool's untouched detached HEAD (zero local-only
commits, no branch ever recorded), no PR.

```
$ hand teardown task-428
warning: herdr tab close failed: herdr tab list --workspace ws-428: exit status 1: {"error":
 {"code":"workspace_not_found","message":"workspace ws-428 not found"}}
id: task-428
result: torn-down
project: hand
kind: ship
outcome: unlanded
detail: worker process exited and no landed work was observed
worktree: returned
exit=0
```

(The Herdr warning is expected and non-fatal: this fixture used fabricated pane/workspace ids since
Herdr release failure was never the question this issue is about, and `releaseHerdr` already treats
it as a warning, not a blocker.)

```
$ treehouse status
1     available    .../hand-a7a4d5/1/hand
2     available    .../hand-a7a4d5/2/hand
```

Both slots back in the pool after the run. `outcome: unlanded`, not `merged` - the completion record
stays honest for an attempt that made no branch and no commits reachable only here.

All scratch resources (both Treehouse leases, the pool, the fleet home, the one-off state-seeding
helper) were released and removed after this run; nothing here touched `projects/`,
`/home/atqa/github/atqamz/hand-dev`, or `/home/atqa/github/atqamz/hand`.

PR #461, opened 2026-08-28
fix(merge): converge on a PR merged outside hand instead of refusing

### LINKED ISSUE #422: A PR merged outside `hand merge` leaves its task permanently unconvergeable: merge refuses, reconcile never looks, deliver records the opposite
## What happens

`hand merge` refuses, permanently, when a task's PR was merged outside `hand`:

```
$ hand merge 408-orient-respects-holds
error: "PR https://github.com/atqamz/hand/pull/414 is already merged"
kind: precondition
exit: 3
help[1]:
  - Nothing changed: this refuses until the state it names is fixed, then the same command runs again
```

The help text names an unsatisfiable precondition. "Already merged" cannot be un-merged, so "then the same command runs again" is never reachable. Every other exit-3 precondition in `hand` describes something the operator can go fix; this one does not.

## Why nothing else recovers it

`hand reconcile` will not observe landing either. `observeLanding` is only reached through `terminalConvergenceCandidate`, which requires the attempt to be `AttemptRunning` **and** its Herdr pane to be absent:

```go
func terminalConvergenceCandidate(attempt state.Attempt, observation reconciliationObservation) bool {
	if attempt.Lifecycle != state.AttemptRunning || attempt.TeardownTerminalAttempt != "" || attempt.Herdr.PaneID == "" {
		return false
	}
	if observation.Herdr.State == herdrOwnershipAbsent {
```

While the worker's pane is alive, reconcile reports `result: healthy / action: keep` and never asks GitHub whether the PR merged. Confirmed on this task after PR #414 merged: reconcile still returned `healthy / keep / idle-unreported`.

So a task whose PR merged out-of-band while its worker is still alive has **no path to a merged lifecycle at all**. `hand merge` refuses, `hand reconcile` does not look, and `hand deliver` records the opposite ("landing it is someone else's call") for work that has already landed.

## Why out-of-band merges are not an operator mistake

They are normal:

- an operator merges from the GitHub web UI;
- a merge queue or `release-please` lands it;
- a stacked PR must go through `pulls/N/merge-async`, which `hand merge` does not use;
- the supervisor merges with `gh pr merge` because that is what every other repo in the fleet uses.

I did the last of those, which is how I found it. But the first three are outside anyone's control.

## Ask

Make `hand merge` converge instead of refuse. If it observes the PR already merged, record the merge - same durable state a `hand merge`-driven merge would have produced - and report it as converged rather than as a precondition failure. That is exactly the "converge durable task state with observed external reality" contract `hand reconcile` already claims.

Second, and separable: let `reconcile` observe landing for a task that has a recorded PR even when its pane is alive. A merged PR is landing evidence regardless of whether the worker is still sitting there. Today the evidence is ignored precisely when it is most current.

Related: atqamz/hand#407 rewrites the `task.PR == ""` arm of `observeLanding`; this issue is about the gate above it that decides whether `observeLanding` runs at all. The two do not collide.

atqamz:
## Summary
- `hand merge`'s already-merged path recorded a precondition ("PR is already merged") that can never be satisfied, so it now converges instead: it records the observed merge with `SetTaskMergeAnnounced`, distinct from `SetTaskMerge`/`MergeExecuted`, and reports `result: converged`.
- `reconcile`'s `terminalConvergenceCandidate` gate required an absent Herdr pane before it would even ask GitHub about a task's PR. `shouldObserveLanding` now asks regardless of pane state whenever a task has a recorded PR, and `decideTerminalConvergence` completes the Attempt on landed evidence pane-alive or not, while unlanded evidence still cannot interrupt a live worker.

Closes atqamz/hand#422

## Test plan
Confirmed atqamz/hand#407 (merged as #415) only rewrote the `task.PR == ""` arm of `observeLanding`; this change is entirely to the gate above it (`terminalConvergenceCandidate` and its callers), which #407 does not touch.

```
$ make lint
go vet ./...
golangci-lint run
0 issues.
go run ./tools/commentlint .

$ make test
... (all packages) ok

$ make e2e
ok  	github.com/atqamz/hand/tests/e2e	67.051s

$ make contract
ok  	github.com/atqamz/hand/tests/contract	0.667s
```

Live drive in a scratch fleet home under `/tmp` (built the real `hand` binary, seeded two Running attempts directly via `internal/state` since a real `hand spawn` needs a live harness/herdr session, and installed a tiny fake `gh` through `hand integration install github/gh --path` so `hand merge`/`hand reconcile` see a real GitHub call resolve, not a Go-test fake):

```
$ hand merge task-1          # PR flipped to MERGED outside hand
id: task-1
result: converged
pr: "https://github.com/example/demo/pull/1"
help[1]:
  - hand observed PR https://github.com/example/demo/pull/1 already merged on GitHub and recorded it; run `hand teardown task-1` to release this task's worktree and pane

$ hand status task-1
task_lifecycle: open
attempt_lifecycle: running
pr: "https://github.com/example/demo/pull/1"
flags: merged-external

$ hand merge task-1          # idempotent rerun, no second `gh pr merge`
id: task-1
result: converged
pr: "https://github.com/example/demo/pull/1"

$ hand reconcile task-2      # attempt running, no Herdr pane recorded, PR still open
result: needs-repair
landing: unlanded
repair_code: herdr-ownership-incomplete

$ hand reconcile task-2      # PR flipped to MERGED outside hand
result: healthy
action: converge-terminal-lifecycle
landing: landed
detail: attempt 2 converged to completed on observed evidence

$ hand status task-2
task_lifecycle: terminal
attempt_lifecycle: completed
```

Cleaned up the scratch fleet afterward (`rm -rf` the temp home, `hand fleet prune --apply` to drop its registry entry). That prune swept 3360 other stale rows already classified `missing` system-wide; both live fleets (`hand-dev`, `hand`) verified untouched and `ready`. Flagging this since I ran `--apply` without first scoping it to only my own entry.

### DIFF
--- a/cmd/merge.go
+++ b/cmd/merge.go
@@ -114,7 +114,7 @@ func runPRMerge(cmd *cobra.Command, home string, t state.Task, method string) er
 		return &ExitError{Err: fmt.Errorf("state of PR %s could not be observed, so hand refuses to merge it: %s", t.PR, observation.Reason()), Code: 3}
 	}
 	if observation.Merged {
-		return &ExitError{Err: fmt.Errorf("PR %s is already merged", t.PR), Code: 3}
+		return convergeAlreadyMergedPR(cmd, home, t)
 	}
 
 	green, err := prChecksGreen(t.PR)
@@ -138,20 +138,8 @@ func runPRMerge(cmd *cobra.Command, home string, t state.Task, method string) er
 	t.MergeExecuted = true
 	t.MergeExecutedAt = mergedAt
 
-	if proj, exists, err := project.Find(home, t.Project); err != nil {
+	if err := syncProjectAfterMerge(cmd, home, t); err != nil {
 		return err
-	} else if exists {
-		releaseProject, err := state.Lock(home, "project:"+proj.Name)
-		if err != nil {
-			return fmt.Errorf("lock project %q: %w", proj.Name, err)
-		}
-		_, syncErr := syncOneProject(home, proj)
-		releaseProject()
-		if syncErr != nil {
-			if _, printErr := fmt.Fprintf(cmd.ErrOrStderr(), "warning: project sync failed: %v\n", syncErr); printErr != nil {
-				return printErr
-			}
-		}
 	}
 
 	var doc axi.Doc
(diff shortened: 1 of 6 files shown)


atqamz:

I went looking for a problem with `SetTaskMergeAnnounced` and found the opposite - worth recording so the next reader does not repeat the search. My concern was that `merge_announced` is also written from the report channel by `SetTaskReportState`, which would have meant recording hand's own GitHub observation into a worker-claim field. It does not: `merge_announced` is already the *observed-externally* field. Its JSON tag is `pr_merged_observed`, `cmd/statusview.go:116` renders it as `merged-external`, and `internal/runtime/pr.go:22` already sets it from `DetectPR`'s GitHub observation. So `announced` versus `executed` is exactly the observed-versus-performed split locked decision 1 asked for, it predates this change, and skipping a new ADR is justified rather than a shortcut.

The gate change is the careful part and it is right. `shouldObserveLanding` opens on a recorded PR regardless of Herdr state, but `decideTerminalConvergence` completes the attempt only on `landingLanded` - unlanded evidence stays behind `terminalConvergenceCandidate`, so "not merged yet" can never interrupt a worker that is still running. That is locked decision 3 honoured precisely: the pane is removed from the *gate* without becoming part of the *determination*. `TestReconcileKeepsLiveWorkerWithUnmergedRecordedPR` and `TestReconcileKeepsLiveWorkerWhenPRStateUnobservable` pin both sides of it, and the unobservable case is the one that most needed pinning.

Confirming atqamz/hand#407's arm is untouched, as asked - it rewrote the `task.PR == ""` branch and this change only reaches tasks with a PR recorded.

One observation, not a request: `hand reconcile` with no id loops every open task, so a fleet with N tasks carrying recorded PRs now makes N GitHub calls where it previously made none for the pane-alive ones. That is fine - reconcile is operator-invoked, not polled, and the watcher does not call it - but it is worth knowing before anyone wires reconcile into a loop.

PR #462, opened 2026-08-28
fix(doctor): report a Treehouse lease held by a Fleet absent from the registry

### LINKED ISSUE #432: doctor: a Treehouse lease held by a Fleet id absent from the registry is silently honoured
## What happens

Treehouse records a lease holder id on each leased pool slot. Nothing cross-checks that id against
hand's Fleet registry, so a slot leased to a Fleet that does not exist reads as legitimately busy:
`hand doctor` says nothing, and the slot is never reclaimed.

Measured on this home 2026-08-26: `/home/atqa/.treehouse/hand-422b27` slot 5 is leased to
`hand:f_167a403f6e12d103a5d310cc10fecedc:340-scout`, and that Fleet id is absent from
`/home/atqa/.secondhand/registry.db` entirely. The slot is therefore permanently unavailable and
nothing reports why.

## Why it is its own issue

This was originally added as an acceptance criterion on atqamz/hand#412
(issuecomment-5423755311, "A Treehouse lease held by a Fleet id absent from the registry is reported,
not silently honoured"). Splitting it out because #412's PR is scoped to metadata soundness -
detecting a slot whose `gitdir:` target is missing, shared, or foreign - which is the corruption
prevention. This is lease-holder liveness, it reaches into the registry rather than into git
metadata, and it interacts with atqamz/hand#413: while the registry holds 3366 junk `/tmp` entries
in state `missing`, "absent from the registry" and "present but dead" are both things a check here
has to distinguish, and #413 has to land first for that distinction to be worth trusting.

## Scope

- Report a leased slot whose holder Fleet id is not in the registry, naming the slot path and the id.
- Distinguish it from a holder that *is* registered but whose Fleet is not `ready` - those want
  different repairs, the same way #412 separates a missing metadata target from a shared one.
- Do not auto-reclaim. A lease is the only record that something believed it owned that directory,
  and a wrong reclaim is worse than a stuck slot. Report, and let the operator decide.
- Do not treat an unparseable holder string as absent; that is a third case and should say so.

## Acceptance

- `hand doctor` reports a lease whose holder id is absent from the registry, naming both.
- A registered-but-not-ready holder is reported distinctly, or deliberately not reported, with the
  choice written down.
- Nothing is reclaimed automatically.
- A test builds a pool with a lease held by an unregistered id and proves it is reported.

Depends on: atqamz/hand#413
Related: atqamz/hand#412, atqamz/hand#403

atqamz:
## Summary

- \`hand doctor\` now classifies every leased Treehouse pool slot's \`lease_holder\` into exactly one of four cases: absent from the Fleet registry, registered but not ready, unparseable, or registered and ready. The first three are reported with their own treatment text naming \`hand fleet\` as the next step; a ready holder stays silent, keeping the common case cheap.
- Report only, never reclaim: nothing here returns, releases, or touches a lease. A registry that cannot be read is reported as its own diagnosis and never mistaken for an absent holder (an absent registry.db - nothing ever registered - is treated as legitimately empty, not unreadable).
- \`worktree.PoolStatus\` and \`worktree.ParseLeaseHolder\` are new: the fix reads \`lease_holder\` from \`treehouse status --json\`, a field nothing in hand consumed before this change (confirmed against the real \`treehouse\` binary while building this - \`lease_id\` is an opaque, treehouse-generated identity, a different field from the \`hand:<fleet-id>:<task-id>\` string).

Closes atqamz/hand#432

## Test plan

- \`make lint\`, \`make test\` (race), \`make e2e\`, \`make contract\` all pass.
- Five focused tests, one per case plus the registry-unreadable case: \`TestDoctorReportsLeaseHolderAbsentFromRegistry\`, \`TestDoctorReportsLeaseHolderRegisteredButNotReady\`, \`TestDoctorReportsUnparseableLeaseHolder\`, \`TestDoctorSaysNothingForLeaseHolderRegisteredAndReady\`, \`TestDoctorReportsRegistryUnreadableRatherThanTreatingHoldersAsAbsent\`; plus \`TestPoolStatusListsEverySlotFromThePoolOutsideEveryFleetHome\` and \`TestParseLeaseHolder\` in \`internal/worktree\`.
- Demonstrated against a scratch fleet home, scratch \`SECONDHAND_HOME\`/registry, and a real \`treehouse\` pool (never the operator's own), leasing four slots with the four holder shapes. Real \`hand doctor\` TOON output:

\`\`\`
findings[4]{line,severity,finding}:
  none,error,"pool slot .../1/demo is leased to task \"340-scout\" of Fleet f_167a403f6e12d103a5d310cc10fecedc, which is absent from the Fleet registry; the lease is not reclaimed automatically, run \`hand fleet\` and decide with the operator"
  none,warning,"pool slot .../2/demo is leased to task \"401-repair\" of Fleet f_84ae2fcea60b9ef578e668309211b6da, which is registered but not ready (missing); the lease is not reclaimed automatically, run \`hand fleet\` to see why"
  none,warning,"pool slot .../3/demo is leased to \"codex-196\", which does not parse as a Hand lease holder (want hand:<fleet-id>:<task-id>); it is not reclaimed automatically, inspect it with \`treehouse status\` before touching the slot"
  none,warning,"routing falls back to legacy defaults without explicit intent: harness \"claude\""
\`\`\`

  Slot 4, leased to this fleet's own registered, ready identity, produced no finding - confirmed stable across three repeat runs.
- The registry-unreadable case is exercised directly at the \`doctorLeaseHolderFindings\` level by \`TestDoctorReportsRegistryUnreadableRatherThanTreatingHoldersAsAbsent\` (a corrupted \`registry.db\`). Note: reaching a corrupt registry.db through the full \`hand doctor\` CLI is masked by \`cmd/root.go\`'s pre-existing \`registry.Preflight\` check, which already refuses any command on a corrupt registry before doctor's own findings run - real output for that path is \`error: "list Fleet locators: file is not a database (26)"\`, a diagnosis about the registry and no holder, consistent with the invariant but via an earlier, unrelated guard.


atqamz:

The missing-registry fix does what I asked and two things I did not.

**Unparseable no longer depends on the registry.** Classifying a holder string as unparseable never needed a registry lookup - the string is malformed regardless of what is registered - and moving that classification ahead of the lookup is right for its own sake, not just as a side effect of the restructure.

**The registry is consulted only when a holder is waiting on the answer.** `if len(toClassify) == 0 { return findings }` means a fleet with no parseable leased holders emits no registry finding at all. That is the correct reading of "one cause, one diagnosis" taken one step further than I stated it: a registry that cannot be consulted is only a problem when something needed consulting it.

Missing and unreadable now read differently and carry different treatments - `hand init` to re-register for the missing case, and a plain "could not be read" for the unreadable one, which is right because those want different repairs. And the unparseable finding tells the operator to inspect with `treehouse status` before touching the slot rather than implying a reclaim, which holds the report-never-reclaim line in the message text as well as the code.

PR #463, opened 2026-08-28
fix(runtime): diagnose a git bundle with no https helper

### LINKED ISSUE #440: fix(runtime): the pinned git bundle ships no remote helpers, so hand project add over https cannot clone
## What happens

`hand project add` with the https URL form README documents fails:

```
$ hand project add https://github.com/atqamz/hand
error: "git clone failed: Cloning into '/home/atqa/github/atqamz/hand-dev/projects/hand'...
warning: templates not found in /dist/git-2.51.0/share/git-core/templates
git: 'remote-https' is not a git command. See 'git --help'.
fatal: remote helper 'https' aborted session
"
kind: general
exit: 1
```

The same source over ssh succeeds:

```
$ hand project add git@github.com:atqamz/hand.git
name: hand
result: added
```

## Cause

The pinned runtime's git bundle contains no remote helpers, and its compiled `--exec-path` points into the build container rather than the installed bundle:

```
$ hand runtime status
git: /home/atqa/.secondhand/runtime/bundles/rd2343fe130ff5ba2-.../git/git
git_version: 2.51.0

$ ls .../git/
git  git-cvsserver  git-receive-pack  git-shell  git-upload-archive  git-upload-pack  gitk  scalar

$ find .../git -name 'git-remote-*'
(nothing)

$ .../git/git --exec-path
/dist/git-2.51.0/libexec/git-core        # does not exist on this machine
```

`git-remote-https` is an external helper, so every https transport operation through the managed git fails. ssh works because git carries that transport internally and needs no helper. The stray `templates not found in /dist/...` warning is the same missing-payload problem showing through.

## Impact

README's Quick start documents `hand project add https://github.com/you/project` as the first of three registration forms, and it cannot work. An operator whose remote is https-only has no working path at all: not `project add`, and not any later fetch or push a worker needs. The failure surfaces as a git error rather than as a runtime defect, so the recovery it suggests is unrelated to the real cause.

## Expected

The pinned bundle carries the remote helpers its documented uses need, with an `--exec-path` that resolves inside the installed bundle, and `hand runtime ensure` / `hand doctor` treat a bundle missing them as unready rather than reporting `ready: true`.

Observed with `runtime_id: rd2343fe130ff5ba2`, `git_version: 2.51.0`, `linux/amd64`, `hand` built from fc9ee9e.

atqamz:
## Summary

- The pinned Git (an external release from `darkvertex/static-git` that hand only consumes, not something this repo builds) ships no `git-remote-*` helpers, so `hand project add https://...` fails with a raw git plumbing error while `hand runtime ensure`/`hand doctor` still report `ready: true`. Confirmed against the real installed bundle that pointing `GIT_EXEC_PATH` at it changes nothing: the helper is absent from the payload, not merely mis-pointed-to.
- `Runtime.SupportsGitTransport` observes `git-remote-<scheme>` next to the managed git binary (no version pinning), and `Store.Status` carries it as `GitHTTPSReady`, kept independent of `Ready` so a fleet home that only clones over ssh never reads as unready with no treatment. `hand runtime status`/`ensure` and `hand doctor` (a non-blocking warning) now name the ssh form as the treatment.
- `hand project add`'s clone path replaces git's "is not a git command" text with that same diagnosis, matched against git's actual failure output rather than guessed from the URL scheme, so a source a git config rewrite resolves without an external helper (as e2e's own fixtures do) is left untouched.

Closes atqamz/hand#440

## Test plan

```
$ make lint
go vet ./...
golangci-lint run
0 issues.
go run ./tools/commentlint .

$ make test
ok  	github.com/atqamz/hand/cmd	...
ok  	github.com/atqamz/hand/internal/toolchain	...
... (all packages ok)

$ make e2e
ok  	github.com/atqamz/hand/tests/e2e	70.628s
```

Real behavior against the currently installed pinned bundle (`HAND_HOME` unset, scratch fleet home):

```
$ hand runtime status
target: linux/amd64
runtime_id: rd2343fe130ff5ba2
ready: true
...
git_https_ready: false
...
help[2]:
  - Run `hand runtime ensure` to install or repair the exact private Git, Treehouse, and Herdr bundle
  - the pinned runtime's Git has no https transport helper (git-remote-https); use the ssh remote form instead, e.g. `hand project add git@host:owner/repo.git`

$ hand doctor
runtime_ready: true
...
ready: true
blocking[0]:
findings[2]{line,severity,finding}:
  none,warning,"routing falls back to legacy defaults without explicit intent: harness \"claude\""
  none,warning,"runtime git cannot clone https:// remotes: the pinned runtime's Git has no https transport helper (git-remote-https); use the ssh remote form instead, e.g. `hand project add git@host:owner/repo.git`"

$ hand project add https://github.com/atqamz/hand
error: "cannot clone https://github.com/atqamz/hand: the pinned runtime's Git has no https transport helper (git-remote-https); use the ssh remote form instead, e.g. `hand project add git@host:owner/repo.git`"
kind: general
exit: 1

$ hand project add git@github.com:atqamz/hand.git
name: hand
result: added
mode: direct-pr
...
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -117,10 +117,13 @@ The target must be an empty or absent directory, or a fleet home `hand init` alr
 ### 3. Add or create a project
 
 ```sh
+hand project add git@github.com:you/project.git
 hand project add https://github.com/you/project
 hand project add ~/work/project
 hand project create new-project
 ```
+
+The currently pinned Git carries no https transport helper, so the ssh form above is the reliable default until a bundle that includes one ships; `hand runtime status` reports this as `git_https_ready`.
 
 Remote sources are cloned under the fleet home and prepared for isolated worker worktrees.
 Local Git sources are adopted once into a separate Fleet-managed clone; Hand never executes in or synchronizes with the original checkout.
@@ -270,7 +273,7 @@ Its internal review, test, documentation, lint, PR, and CI stages remain owned b
 Choose a mode when registering a project:
 
 ```sh
-hand project add https://github.com/you/project --mode direct-pr
+hand project add git@github.com:you/project.git --mode direct-pr
 ```
 
 Add an existing local Git project or create a new managed project with the local-only mode:
--- a/cmd/doctor.go
+++ b/cmd/doctor.go
@@ -121,6 +121,7 @@ func newDoctorCmd(info selfupdate.BuildInfo) *cobra.Command {
 			if err != nil {
 				return err
 			}
+			findings = append(findings, runtimeHTTPSFindings(runtimeStatus)...)
 			herdrSession := observeCurrentHerdrSession(cmd.Context(), fleetHome)
 			tools := doctorManagedTools(runtimeStatus, projects)
 			integrations, err := integration.DefaultStore().List()
@@ -182,12 +183,23 @@ func doctorRuntimeStatus() (toolchain.Status, error) {
 	if legacyDoctorCompatibility() && !status.Ready && onPath("git") && onPath("treehouse") && onPath("herdr") {
 		status.Ready = true
 		status.Reason = "test-only legacy tool fixture"
+		status.GitHTTPSReady = true
 	}
 	return status, nil
 }
 
 func legacyDoctorCompatibility() bool {
 	return legacyDoctorCompat
+}
+
+// A missing https helper is a warning, not blocking: the bundle is otherwise intact and ssh
+// keeps working, so a fleet home that only clones over ssh must not read as unready with no
+// treatment (hand#440). A runtime that fails Ready already has its own "runtime" blocking entry.
+func runtimeHTTPSFindings(status toolchain.Status) []doctorFinding {
+	if !status.Ready || status.GitHTTPSReady {
+		return nil
+	}
+	return []doctorFinding{{Severity: doctorWarning, Text: "runtime git cannot clone https:// remotes: " + gitHTTPSTreatment("https")}}
 }
 
 func doctorFindings(fleetHome string, histories []state.TaskHistory) ([]doctorFinding, error) {
--- a/cmd/doctor_test.go
+++ b/cmd/doctor_test.go
@@ -19,7 +19,42 @@ import (
 	"github.com/atqamz/hand/internal/skill"
 	"github.com/atqamz/hand/internal/state"
 	"github.com/atqamz/hand/internal/supervision"
+	"github.com/atqamz/hand/internal/toolchain"
 )
+
+// The doctor half of hand#440: a bundle that is otherwise intact but cannot clone https:// is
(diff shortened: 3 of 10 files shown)


atqamz:
Reviewed. The design is right and one detail in it is better than what I asked for. One addition before this lands.

**Accepted, and confirming the judgment call explicitly.** Keeping `GitHTTPSReady` independent of `Ready` is correct for the reason you gave - `Ready` also gates treehouse and herdr readiness and doctor's blocking list, so flipping it would make every fleet home on the pinned bundle read as unready with no treatment available. That was the escalation case in the brief and a transport-specific field plus a non-blocking warning is a better answer than either extreme. `runtimeStatusHelp` keeping the generic repair line first and noting that `hand runtime ensure` reinstalls the same helper-less Git is the kind of detail that stops an operator wasting a cycle.

**The detail I did not ask for and should have:** diagnosing the clone failure by matching git's actual failure text rather than guessing from the URL scheme. A scheme-based guess would have misreported any source that a `git config` `insteadOf` rewrite resolves without an external helper - which the e2e fixtures themselves rely on. `TestDiagnoseCloneFailureLeavesUnrelatedGitErrorsUnchanged` pins it.

**Q1 answered:** the pinned Git is an external release from `darkvertex/static-git` that hand only consumes. So the payload half of the issue's "Expected" is not this repository's to satisfy, and this PR correctly delivers only the half hand owns.

**One addition, then this can land.** `README.md:120` documents

```
hand project add https://github.com/you/project
```

as the first of three registration forms, and `README.md:273` repeats it with `--mode direct-pr`. On the pinned bundle that cannot work, and this PR makes the failure honest without making the documentation honest. The issue's own Impact section calls this out: "an operator whose remote is https-only has no working path at all".

Update those so a reader reaches for the ssh form first, with a brief note that the pinned runtime's Git carries no https transport helper and `hand runtime status` reports `git_https_ready`. Keep it short - a sentence, not a section - and do not remove the https form, since it will work again once the bundle carries the helper.

Worth adding an INV row only if you think the docs claim is one; I do not, so use your judgement.

atqamz:
Updated README.md at both spots: the Quick Start and delivery-mode registration examples now lead with the ssh form, https stays documented, and a one-sentence note points at `hand runtime status`'s `git_https_ready`. Pushed as da0d0f0. Not merging.

atqamz:
README change reviewed. Accepted, nothing further.

Both sites are right and the shape is what I wanted: the ssh form leads at `README.md:120` and at the delivery-mode example, https stays documented rather than being removed, and the note is one sentence naming `git_https_ready` so a reader can check the condition themselves rather than taking the README's word for it. "until a bundle that includes one ships" is the right framing too - it dates the limitation to the bundle rather than to hand.

Merging once CI is green.

PR #465, opened 2026-08-28
fix(watcher): stop failing an arm when a task is torn down mid-probe

### LINKED ISSUE #455: watch: tearing a task down while the wake bridge arms fails the whole arm, blocking the supervisor's Stop hook
## What happened

Observed live while supervising this milestone. I ran `hand teardown 410-lock-holder-misattribution` after merging its PR; the Supervisor Harness's Stop hook fired immediately after, and the wake bridge refused:

```
error: "hand supervisor wake bridge failed monitoring: could not arm: 410-lock-holder-misattribution: herdr pane get w4:p5: exit status 1: {\"error\":{\"code\":\"pane_not_found\",\"message\":\"pane w4:p5 not found\"},\"id\":\"cli:pane:get\"}; run `hand doctor` in the fleet home"
kind: usage
exit: 2
```

Seven other tasks were running and healthy. `hand doctor` immediately afterwards reported `ready: true` with nothing to fix, and the next `hand orient` listed six monitor targets with `410-lock-holder-misattribution` correctly gone. Nothing was actually wrong by the time the message asked me to look.

## Mechanism

`armProbe` (`internal/watcher/watcher.go:296-316`) walks the task histories and probes each active attempt's pane:

```go
if history.ActiveAttempt.Lifecycle == state.AttemptProvisioning {
    continue
}
if _, err := client.PaneGetContext(ctx, history.ActiveAttempt.Herdr.PaneID); err != nil {
    done <- fmt.Errorf("%w: %s: %v", ErrArmFailed, history.Task.ID, err)
    return
}
```

A single unprobeable pane returns immediately and fails the whole arm. That is deliberate, and the reason is stated at `internal/watcher/watcher.go:113-114`:

> `ErrArmFailed` names the one task whose pane could not be probed at arm time, so an exit from `RunUntilEvent` always means the whole fleet was actually watched.

The all-or-nothing invariant is right. What is wrong is treating *this* pane as a failure to watch: the task was torn down, on purpose, by the supervisor, moments earlier. The history list was read before the teardown committed, so a task that no longer needs watching was still in the arm set. `AttemptProvisioning` is already skipped one line above for the mirror-image reason - a pane that does not exist yet is not a pane that cannot be reached.

## Why it is worth fixing rather than tolerating

The blast radius is out of proportion to the cause. A supervisor that merges a PR and tears the task down - the normal end of every task - can have its next Stop hook fail with `exit: 2`, and the remedy the message names (`hand doctor`) reports nothing, because by then there genuinely is nothing. That is a diagnosis with no reachable treatment, which `docs/adr/every-diagnosis-names-a-reachable-treatment.md` exists to prevent.

It also fails *closed on the wrong thing*: the seven live workers that did need watching went unwatched because one dead pane was in the list.

## Fix direction, not settled

- Skip an attempt whose lifecycle is terminal, the same way `AttemptProvisioning` is skipped, so a torn-down task is never in the arm set. This is the narrow fix and probably the right one.
- Re-read the task's lifecycle when `PaneGetContext` returns `pane_not_found` specifically, and treat "gone because the task is gone" as a skip rather than a failure. Distinguishes the race from a genuinely lost pane.
- Do not weaken the all-or-nothing invariant. A pane that vanished while its task is still open is a real arm failure and must stay one.

## Acceptance

- Tearing a task down concurrently with an arm does not fail the arm.
- A pane that is missing while its task is still open still fails the arm, still names the task.
- No refusal in the wake-bridge path names `hand doctor` as the treatment for a condition `hand doctor` cannot see.

Observed on `3334151`, `herdr_version: 0.8.2`, linux/amd64.

atqamz:
## Summary

- `probeAllTasks` (armProbe) now skips an attempt already carrying teardown's herdr-release mark, and re-reads a task fresh on a `pane_not_found` probe failure before declaring `ErrArmFailed` - closing the race where a concurrent `hand teardown` commits between the histories snapshot and this task's probe turn.
- The Claude wake-bridge's arm-failed refusal now points at `hand status` instead of `hand doctor`, which never inspects pane reachability and could never see the condition it was telling the operator to check.

A static check on the attempt's own `Lifecycle` field (the issue's other suggested direction) cannot close this race and is not included: `ListOpenHistories` only returns tasks whose lifecycle is still `open`, and `TerminalizeTaskAndAttempt` commits both the task and the attempt terminal in one transaction, so a terminal `Lifecycle` can never appear in that snapshot. Only a fresh re-read, taken after the probe has already failed, can see what committed since the snapshot was read. This is explained in the commit message and in code comments.

Closes atqamz/hand#455

## Test plan

```
$ make lint
go vet ./...
golangci-lint run
0 issues.
go run ./tools/commentlint .

$ make test
...
ok  	github.com/atqamz/hand/internal/watcher	60.699s
ok  	github.com/atqamz/hand/cmd	138.407s
... (all packages ok)

$ make e2e
ok  	github.com/atqamz/hand/tests/e2e	102.271s
```

New focused tests in `internal/watcher/watcher_test.go`, all passing:
- `TestProbeAllTasksSkipsATornDownAttemptWithoutProbingItsPane` - an attempt already marked mid-teardown in the histories snapshot is never probed.
- `TestProbeAllTasksClosesTheRaceWhenATeardownCommitsAfterTheHistoriesSnapshot` - the actual reported race: a stale snapshot with a genuinely open, non-terminal attempt, and a teardown mark committed (via the same primitive `hand teardown` uses) after the snapshot but before the probe. Arms successfully and the remaining task is still probed.
- `TestProbeAllTasksFailsWhenAnOpenTasksPaneIsGoneWithoutATeardown` - the re-read must not swallow a genuine failure: still fails, still names the task.
- `TestProbeAllTasksArmsForAnOpenTaskWithALivePane`, `TestProbeAllTasksSkipsAProvisioningAttemptWithoutProbingItsPane` - unchanged branches.

Also manually verified red-before/green-after against the real compiled `hand` binary in a scratch fleet home built by `t.TempDir()` (outside this worktree, per atqamz/hand#413): a two-task fleet with task-2 marked mid-teardown and its pane answering `pane_not_found` failed `hand watch --until-event` with `kind: arm-failed` before this change, and passes - delivering task-1's event - after it. That verification script was throwaway and is not part of this PR.

`docs/testing-invariants.md` gained `INV-WATCH-9`, `INV-WATCH-10`, and `INV-DIAG-4` for the new invariants.

### DIFF
--- a/cmd/supervision.go
+++ b/cmd/supervision.go
@@ -201,7 +201,9 @@ func newClaudeStopCmd() *cobra.Command {
 				if markErr := ledger.MarkBridgeError(harness.Claude, waitErr.Error()); markErr != nil {
 					return markErr
 				}
-				return &ExitError{Err: fmt.Errorf("hand supervisor wake bridge failed monitoring: %v; run `hand doctor` in the fleet home", waitErr), Code: 2}
+				// `hand doctor` never inspects pane reachability, so it cannot see this condition -
+				// `hand status` is what actually observes an unreachable pane (atqamz/hand#455).
+				return &ExitError{Err: fmt.Errorf("hand supervisor wake bridge failed monitoring: %v; run `hand status` in the fleet home to see which task's pane is unreachable", waitErr), Code: 2}
 			default:
 				return waitErr
 			}
--- a/docs/testing-invariants.md
+++ b/docs/testing-invariants.md
@@ -159,7 +159,8 @@ Source: [One watcher per fleet home, guarded by an flock](adr/one-watcher-per-fl
 [Watcher takeover is generation-attributed](adr/watcher-takeover-is-generation-attributed.md),
 [The until-event exit is the delivery](adr/the-until-event-exit-is-the-delivery.md),
 [Arming a watch observes before it waits](adr/arming-a-watch-observes-before-it-waits.md),
-[A contended refusal names its recorded holder](adr/a-contended-refusal-names-its-recorded-holder.md).
+[A contended refusal names its recorded holder](adr/a-contended-refusal-names-its-recorded-holder.md),
+`internal/watcher/watcher.go` (`probeAllTasks`, `attemptStillNeedsArm`; atqamz/hand#455).
 
 | id | invariant | layer | coverage |
 |---|---|---|---|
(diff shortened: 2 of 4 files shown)


atqamz:

**You refuted my preference with evidence, which is what I asked for.** The brief said to prefer the static lifecycle check and fall back to the not-found re-read only if direction 1 did not close the window. You established that direction 1 cannot close it *at all*: `ListOpenHistories` returns only open tasks and `TerminalizeTaskAndAttempt` commits task and attempt terminal atomically, so a terminal Lifecycle can never appear in the snapshot. Only the fresh re-read closes it.

What makes this better than simply doing the fallback is that you did not then ship the unreachable check as decoration. The skip is on `TeardownHerdrState != ""` - an attempt already marked mid-teardown, which *is* reachable in a snapshot - so both paths are load-bearing and cover different windows: already-marked at snapshot time, and committed between snapshot and probe. INV-WATCH-9 names both, which is the right shape for a row.

**The detail I would have missed:** `attemptStillNeedsArm` checks `history.ActiveAttempt.ID != probedAttemptID`, not just lifecycle and teardown state. That stops a `pane_not_found` being forgiven for a task that has since been reopened onto a *new* attempt - a genuinely missing pane on live work, forgiven because the old attempt went away. Nothing in the issue or the brief pointed at that.

**INV-WATCH-10 is the row that matters most** and I am glad it exists as its own line rather than a clause: a pane missing for an open task not being torn down still fails the arm and still names the task, *through the same not-found path INV-WATCH-9 forgives*. That is precisely where a fix like this goes wrong, and it is pinned.

The wake-bridge refusal naming `hand status` instead of `hand doctor` is locked decision 3 honoured, and INV-DIAG-4 records why: `hand status` checks pane reachability and `hand doctor` does not, which is exactly why the original message sent me to a command that reported `ready: true` while the arm was failing.

Issue #459, opened 2026-08-28
hand send cannot confirm an Enter submission on codex: the accepted message stays on screen as history, so composerRetains never clears


atqamz:
Split out of atqamz/hand#420 during its fix. Found by live testing against a real codex worker, before anything shipped.

## The gap

atqamz/hand#420's fix confirms a send by reading the pane back and checking whether the text hand sent is still sitting in the composer. On the Tab path that works: `codexQueuedMarker` gives `composerRetains` a truncation point, so only text before "Queued follow-up inputs" counts, and a message that was genuinely queued never reads as retained.

**The Enter path has no equivalent, and for codex it can essentially never succeed.**

Observed live: a normal instruction was sent to an idle codex worker. `chooseSubmitKey` read the pane, saw no queue marker, and correctly chose Enter. The send genuinely succeeded - codex accepted it as a new turn and responded. But codex echoes the accepted message as a `>`-prefixed history line that stays visible indefinitely, and `composerRetains` has no way to tell "still sitting unsent at the live composer" from "already accepted, now history". The bounded 10s poll never cleared.

This is not a flaky signal. It is wrong by construction for this harness: after *every* successful Enter send to codex, the text is still on screen.

## Why it was not shipped

Two compounding failures, and the second is the serious one:

1. **False negative.** A send that succeeded reports `not-submitted`, landing in the operator-attention channel for no reason.
2. **Duplicate delivery.** The confirmation failure triggered a retry, which resent the identical text and Enter, and codex accepted it as a *second, separate turn*. Confirmed by a direct pane read showing both `>`-prefixed copies with a real codex response after each. That is a behaviour regression, not a labelling one - a worker told twice may act twice.

So the fix for atqamz/hand#420 landed as atqamz/hand#426 only: harness-aware submit-key selection, the pane-read plumbing, and confirmation on the queued/Tab path where truncation makes it sound. Enter-path confirmation and the retry were both held back.

## What is not affected

`SendUncertain`, the not-submitted-before-acceptance path, retry-count bounding (exactly one retry occurred, not a loop), and the Tab path all behaved as designed. And the held-back behaviour fails toward "needs attention", never toward a false `submitted` - so shipping without it leaves atqamz/hand#420's original defect exactly as it was, no worse.

## Fix directions, not settled

Three candidates, none implemented or validated:

[...]

Issue #420, opened 2026-08-26
hand send reports `send_state: submitted` with a character count when zero characters reached the pane


atqamz:
## What happens

`hand send` reports a successful delivery while roughly half the message is missing from the composer
and Enter never submitted it. Five for five today, across two projects and five different workers.

Each send returned:

```
chars: 3858
help[1]:
  - The terminal pane accepted the message and Enter; run `hand status <id>` to read what the worker does with it
```

Reading the panes afterwards with `herdr pane read` shows, in every case, the message sitting at the
`›` composer prompt, **ending mid-sentence**, with `agent_status: idle` and no new worker report. The
worker never saw any of it.

## Measured

| pane | task | bytes sent | bytes in composer (approx) | kept |
| --- | --- | --- | --- | --- |
| `wK:p6` | 406-brief-report-channel | 5534 | ~3023 | ~54% |
| `wK:p8` | 407-reconcile-landing-evidence | 6028 | ~3021 | ~50% |
| `wK:p9` | 408-orient-respects-holds | 3859 | ~2028 | ~52% |
| `wM:p9` | 6-consumer-fixture | 3688 | ~2018 | ~54% |
| `wM:pA` | 6-adoption-docs | 2238 | ~2031 | ~90% |

The composer byte counts are approximate - I derived them from `herdr pane read` output, which
collapses wrapped lines - so treat the percentages as indicative. What is exact and not
interpretation-dependent: every composer's text ends mid-word or mid-sentence, and every message is
still unsent.

Two things stand out. The retained sizes cluster at ~2020 and ~3020 rather than scaling with input,
and the loss is roughly proportional (~50%) rather than a fixed ceiling. That combination looks like
chunked writes where some chunks are dropped - about every other one - rather than a length limit.
The smallest message (2238 bytes) appears to have arrived whole, so there is likely a threshold below
which this does not happen.

All five workers run harness `codex`, model `gpt-5.6-luna`, effort `high`. The harness builder already
sets `-c disable_paste_burst=true` for exactly this class of problem, so whatever is dropping text is
not the paste-burst path it was added to defeat, or that flag is no longer sufficient.

## Why this is severe

This is the supervisor's only channel to a running worker, and it fails while reporting success.

Concretely, today: three code reviews with confirmed P1 findings and two dispatch decisions resolving
a real scope collision went nowhere, and `hand status` / `hand orient` showed nothing wrong - the tasks
[...]

PR #466, opened 2026-08-28
fix(steering): send Tab, not Enter, to a busy codex pane

### LINKED ISSUE #426: hand send reports submitted after the text arrives but the Enter never submits it
## Summary

`hand send` reported success, the full text did reach the composer, and the message was still
never submitted. The worker stayed `done` with the whole brief sitting in its input buffer. A
single `herdr agent send-keys <pane> enter` submitted it immediately.

Distinct from atqamz/hand#420, where zero characters arrive. Here all 2274 arrived; only the
Enter had no effect, and `send_reason` asserts it was accepted.

## Observed

```
$ hand send 6-consumer-fixture --file /tmp/resume-6cf.md
id: 6-consumer-fixture
result: sent
send_id: 95
attempt: 76
send_state: submitted
chars: 2274

$ herdr --session hand-f_... agent get wM:p9
"agent_status":"done"
"state_change_seq":409
```

The pane read showed the message body rendered in the composer, indented, ending mid-sentence at
the point the viewport cut off - not in the transcript. `agent_status` stayed `done` and
`state_change_seq` did not move. `hand status` recorded `send_state: submitted`,
`send_reason: text-and-enter-accepted`.

Then:

```
$ herdr --session hand-f_... agent send-keys wM:p9 enter
{"id":"cli:agent:send-keys","result":{"type":"ok"}}
```

`agent_status` went to `working` and the worker's first line was
`working: resuming from the parked state`. Nothing else changed in between.

## Why it matters

The supervisor has no way to tell this apart from a delivered message without reading the pane
out of band, and `send_reason: text-and-enter-accepted` actively says the Enter landed. A brief
that silently sits in a composer looks exactly like a worker that received its instructions and
chose to do nothing - which is how this was originally misread as a mute worker.

## Where to look

The submit path presumably writes the text, then sends Enter, then checks something that
reported "accepted". Whatever that check is, it passed while the composer kept the draft. Two
candidates: the Enter is delivered before the multi-line paste has settled in the TUI, so the
composer treats it as part of the paste; or acceptance is inferred from the write succeeding
rather than from observing the pane leave its pre-send state.

## Acceptance

- A send confirms submission by observing the pane actually leave its pre-send state (or the
  transcript gain the message), not by the keystroke write returning without error.
- When that observation fails, `send_state` says so rather than `submitted`, and the recorded
  `send_reason` does not claim the Enter was accepted.
- A test drives a composer that keeps a multi-line draft after one Enter and proves the send is
  not reported as submitted.

Related: atqamz/hand#420, atqamz/hand#419

atqamz:
## Summary

- `hand send` now reads a codex pane before submitting and sends `Tab` instead of `Enter` when the pane shows codex's own `tab to queue message` text (`chooseSubmitKey`) - codex refuses Enter mid-turn and queues the keystroke instead, which is what atqamz/hand#426 reported as a silent `text-and-enter-accepted` false claim.
- That Tab path is then confirmed, not assumed: `composerConfirms` reads the composer back and checks whether a recognizable fragment of the sent message is still there (excluding codex's own echo under `codexQueuedMarker`), landing on `not-submitted`/`uncertain` rather than `submitted` when it can't verify. Enter is untouched - still claim-based on two successful RPCs, exactly as before this change - because live testing found no marker there distinguishes an accepted message's own history line from text still sitting unsent; that gap is atqamz/hand#420, staying open for a follow-up.
- `docs/testing-invariants.md` and a new ADR (`codex-tab-queue-submission-is-confirmed-by-composer-content.md`) scope every confirmation claim to the Tab/codex-queue path only, and record why Enter-path confirmation and a retry-on-no-confirm were both tried, live-tested, and dropped (the retry doubled a real delivery, not just a label).

Closes atqamz/hand#426

## Test plan

```
$ make lint
go vet ./...
golangci-lint run
0 issues.
go run ./tools/commentlint .

$ make test
ok  	github.com/atqamz/hand/cmd	136.957s
ok  	github.com/atqamz/hand/internal/steering	9.725s
ok  	github.com/atqamz/hand/internal/store	25.209s
ok  	github.com/atqamz/hand/internal/watcher	60.067s
(all packages ok)

$ make e2e
ok  	github.com/atqamz/hand/tests/e2e	84.582s

$ make contract
go test -tags=contract -count=1 -timeout=10m ./tests/contract/...
ok  	github.com/atqamz/hand/tests/contract	0.511s
```

Both `chooseSubmitKey`'s Tab selection and `composerConfirms`'s Tab-path confirmation were live-verified earlier in this task against a real busy codex worker over two independent sends; the Enter-path confirmation this PR does *not* ship was also live-tested and found to misreport an ordinary send as not-submitted, then to duplicate a message when retried on top of that - full mechanism and evidence in the new ADR's Context and Rejected alternatives sections.


atqamz:

The split is exactly as directed and the parts held back are held back cleanly - Enter reverts to `text-and-enter-accepted` with no pane read after, Tab confirms via composer content with the `codexQueuedMarker` truncation and lands `text-and-tab-queued`, and the retry is gone entirely. `Closes atqamz/hand#426` only, with #420 left open.

Three things worth saying.

**The ADR keeps the negative result in the tree, not only on an issue.** Recording that Enter-path confirmation was built, live-tested and reverted - with the reason and the two undeveloped directions - means the next person to reach for it finds out it was already tried before they spend the same day on it. That is worth more than the code that was removed.

**Your one-line justification for dropping the retry is better than the one I gave you:** "a retry with no working confirmation on that path could only duplicate a message that had already landed, never recover one that had not." I argued it from harm; that argues it from futility, which is the stronger form.

**`SendReasonComposerRetainsMessage` stays reachable**, via a Tab-path queue that fails to confirm, so the new vocabulary is not dead. I checked because vocabulary nothing can produce was the specific thing I told you not to land.

The `PaneReadUnwrapped` widening into `limitPane` and its test fake is the minimum needed and the fake's comment says why it is never reached. Fine.

One note for whoever picks up atqamz/hand#420: I corrected that issue's sibling atqamz/hand#459 after your follow-up on probe-3. My original wording there said Enter-path confirmation "can essentially never succeed… by construction", and your reason-code check disproved it - probe-3 confirmed clean on the first read. The mechanism is a race, not a static property, and #459 now says so.

PR #467, opened 2026-08-28
feat(cmd): record a PR delivered to a sibling repository with hand pr --cross-repo

### LINKED ISSUE #423: hand pr cannot record a PR delivered to a sibling repository, and the only offered escape misrepresents the project topology

atqamz:
## Summary

- `hand pr` gains an explicit `--cross-repo` flag paired with a required `--reason`, waiving exactly the repo-match check when a PR genuinely landed in a repository other than the project's own or its declared upstream - never by extending `hand project upstream`, which would write false fork topology into the registry.
- The reason lands in a new `task.pr_cross_repo_reason` column (empty for every ordinary PR), and `hand status` marks a cross-repo record distinctly wherever it renders a PR: the `pr` field, the fleet `flags` cell, and `--json`.
- Neither `hand pr`'s own branch-based re-detection nor the watcher's auto-recording from a worker's report line ever sets the opt-in - a cross-repo delivery is asserted by an operator, never discovered. See `docs/adr/cross-repo-pr-delivery-is-an-explicit-opt-in.md` for the full rationale and rejected alternatives.

Closes atqamz/hand#423

## Test plan

```
$ make lint
go vet ./...
golangci-lint run
0 issues.
go run ./tools/commentlint .

$ make test
ok  	github.com/atqamz/hand/cmd	...
ok  	github.com/atqamz/hand/internal/project	...
ok  	github.com/atqamz/hand/internal/runtime	...
ok  	github.com/atqamz/hand/internal/store	...
ok  	github.com/atqamz/hand/internal/watcher	...
(all packages ok)

$ make e2e
ok  	github.com/atqamz/hand/tests/e2e	66.671s
```

Focused new/updated tests: `cmd/pr_test.go` (refusal names `--cross-repo`, both flags required together, cross-repo record + reason, existence check still enforced, write-once still holds, fork-upstream path unaffected), `cmd/status_test.go` (`TestStatusRendersCrossRepoPRDistinctlyEverywhere`), `tests/e2e/pr_test.go` (`TestPRCommandCrossRepo`), `tests/e2e/upstream_deliver_test.go` (updated refusal-text assertion).

Demonstrated both paths for real, in a scratch fleet home built in `/tmp` (not inside a Treehouse worktree, per atqamz/hand#413) against real GitHub state - project `yes2infra` and the real PR `yes2games/butler#92` from the motivating case:

```
$ hand pr butler-broker-allowlist-rujak-release https://github.com/yes2games/butler/pull/92
error: "PR https://github.com/yes2games/butler/pull/92 belongs to yes2games/butler, not project
        yes2infra's repo (yes2games/yes2infra); pass --cross-repo with --reason on `hand pr` if
        this is a deliberate delivery elsewhere"
kind: precondition
exit: 3

$ hand pr butler-broker-allowlist-rujak-release --cross-repo --reason "yes2infra#435 deleted
    pulumi/worker/config/broker.json and moved the Worker to yes2games/butler before this task
    started; the file, BrokerEntry shape and vitest gate moved verbatim (ADR-0015)" \
    https://github.com/yes2games/butler/pull/92
id: butler-broker-allowlist-rujak-release
result: recorded
pr: "https://github.com/yes2games/butler/pull/92"
cross_repo_reason: "yes2infra#435 deleted pulumi/worker/config/broker.json and moved the Worker
    to yes2games/butler before this task started; the file, BrokerEntry shape and vitest gate
    moved verbatim (ADR-0015)"

$ hand status
tasks[1]{id,state,reported,age,flags}:
  butler-broker-allowlist-rujak-release,unknown,none,just now,cross-repo
```

### DIFF
--- a/cmd/pr.go
+++ b/cmd/pr.go
@@ -13,6 +13,9 @@ import (
 )
 
 func newPRCmd() *cobra.Command {
+	var crossRepo bool
+	var reason string
+
 	cmd := &cobra.Command{
 		Use:   "pr <id> <url>",
 		Short: "Record a task's pull request URL",
@@ -21,6 +24,12 @@ func newPRCmd() *cobra.Command {
 			id, url := args[0], args[1]
 			if !state.ValidatePRURL(url) {
 				return &ExitError{Err: fmt.Errorf("invalid PR URL %q: must match https://github.com/<owner>/<repo>/pull/<number>", url), Code: 2}
+			}
+			if crossRepo && reason == "" {
+				return &ExitError{Err: fmt.Errorf("--cross-repo requires --reason describing the deliberate delivery elsewhere"), Code: 2}
+			}
+			if !crossRepo && reason != "" {
+				return &ExitError{Err: fmt.Errorf("--reason applies only to a --cross-repo record"), Code: 2}
 			}
 			fleetHome, err := home.Resolve()
 			if err != nil {
@@ -35,7 +44,7 @@ func newPRCmd() *cobra.Command {
 			if err != nil {
 				return asPrecondition(err)
 			}
-			task, reconcile, err := recordPR(cmd.Context(), fleetHome, task, url)
+			task, reconcile, err := recordPR(cmd.Context(), fleetHome, task, url, crossRepo, reason)
 			if err != nil {
 				return err
 			}
@@ -50,6 +59,9 @@ func newPRCmd() *cobra.Command {
 			doc.Field("id", task.ID)
 			doc.Field("result", result)
 			doc.Field("pr", url)
+			if task.PRCrossRepoReason != "" {
+				doc.Field("cross_repo_reason", task.PRCrossRepoReason)
+			}
 			// A torn-down task has no active attempt for hand merge to act on (atqamz/hand#424): naming
 			// it here would suggest the task is live again, which recording a PR on it must never do.
 			if task.Lifecycle != state.TaskTerminal {
@@ -58,11 +70,13 @@ func newPRCmd() *cobra.Command {
 			return doc.Render(cmd.OutOrStdout())
 		},
 	}
+	cmd.Flags().BoolVar(&crossRepo, "cross-repo", false, "record a PR in a repository other than the project's own or its declared upstream, deliberately")
+	cmd.Flags().StringVar(&reason, "reason", "", "why this PR landed in a different repository, required with --cross-repo")
 	return cmd
 }
 
-func recordPR(ctx context.Context, homeDir string, task state.Task, url string) (state.Task, bool, error) {
-	updated, reconcile, err := runtime.RecordPR(ctx, homeDir, task, url)
+func recordPR(ctx context.Context, homeDir string, task state.Task, url string, crossRepo bool, reason string) (state.Task, bool, error) {
+	updated, reconcile, err := runtime.RecordPR(ctx, homeDir, task, url, crossRepo, reason)
 	if err != nil {
 		return task, false, asPrecondition(err)
 	}
--- a/cmd/pr_test.go
+++ b/cmd/pr_test.go
@@ -446,6 +446,173 @@ func TestPRNotFoundOnUpstreamNamesTheUpstreamRepo(t *testing.T) {
 	}
 }
 
+// atqamz/hand#423: the refusal for a sibling repo has to point at the real escape (--cross-repo),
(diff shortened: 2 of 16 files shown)


atqamz:
Reviewed. Accepted, no changes requested. Merging once CI is green, though see the ordering note at the end.

All four locked decisions are honoured, and one boundary I did not state is the best thing in the change.

**`autoRecordPR` and `DetectPR` both hardcode `crossRepo: false`.** Hand never infers a cross-repo delivery - not from a worker's report line, not from re-detecting a branch. The ADR's reason for the second is sharper than "be conservative": re-detection can only ever find a PR on the project's own repo or its declared upstream, because that is where hand pushed the branch. So an inferred cross-repo record would be evidence of a bug, not a discovery. Locking the flag to the operator's assertion is exactly the "not by accident, rather than not at all" line the issue asked for.

**The `--reason` guard runs both ways.** `--cross-repo` without `--reason` refuses, and `--reason` without `--cross-repo` refuses too. I only asked for the first. The second matters more than it looks - a stray `--reason` on a same-repo record would imply a deviation that did not happen, and the record is the thing a later reader reconstructs the decision from.

**The `cross-repo` flag in `taskFlags`** satisfies decision 4, and gating it on `PRCrossRepoReason != ""` rather than a separate boolean means the flag and its explanation cannot drift apart.

Rejecting `hand project upstream` in the ADR with the right distinction - a standing topological fact versus a one-off event - records why the obvious escape was wrong, which is what will stop someone re-proposing it.

**Ordering note.** PR #466 (atqamz/hand#426) is ahead of you in CI and also touches `internal/watcher/watcher.go`, and PR #465 (atqamz/hand#455) touches `probeAllTasks` in that same file. Different functions, so git will probably cope. The more likely conflict is `internal/store/schemaversion.go`: your new migration appends to `migrations`, and anything else adding one lands in the same place. If either merges first, rebase and make sure your migration keeps a distinct index rather than being merged into someone else's slot.

PR #486, opened 2026-08-28
diagnose stale flake.nix vendorHash on CI failure

### LINKED ISSUE #483: Every Dependabot bump fails the Nix build, because flake.nix's vendorHash is not refreshed with go.mod
## What

`flake.nix` pins `vendorHash`, the fixed-output hash of what `go mod vendor` produces. Any change to `go.mod` or `go.sum` invalidates it, and the `Nix build (x86_64-linux)` job then fails on a hash mismatch:

```
error: hash mismatch in fixed-output derivation '...-hand-0.6.0-go-modules.drv':
         specified: sha256-J4BQ8DVzZaG+KMBa61tMDuoOtcH3AgcTYwr97mjLitU=
            got:    sha256-oxRfw0HIVStGdzZnzdKdFAIX+ntWOjsormX37rz3tag=
```

Dependabot does not know `flake.nix` exists, so every dependency PR it opens is red on that job from the moment it is created, and stays red until a human rebases the branch and recomputes the hash by hand.

## Observed

atqamz/hand#445 (`modernc.org/sqlite` 1.56.0 to 1.57.0) and atqamz/hand#446 (`go.yaml.in/yaml/v3` 3.0.4 to 3.0.5) both failed this way. Both needed a manual rebase onto `main` plus a recomputed hash before they could merge.

atqamz/hand#446 needed it **twice**: the hash computed against `main` was invalidated the moment atqamz/hand#445 merged, because both PRs change `go.mod`. Two dependency bumps in flight at once cost three hash computations, and the second one is only discoverable after the first merges.

It also hits ordinary work. atqamz/hand#442's phase 1 added `pgregory.net/rapid` and had to update the same hash; the worker caught it before pushing, but only because it thought to run `nix build` locally.

## Why fix it rather than absorb it

The manual step is small and entirely mechanical, which is what makes it worth automating: it is invisible until CI is red, the fix requires a Nix toolchain the PR author may not have, and the cost grows quadratically with the number of dependency PRs open at once. It also trains everyone to see a red Nix job on a dependency PR as normal, which is exactly when a real Nix build failure gets waved through.

## Directions

Whoever picks this up should weigh at least these:

- **A CI step that recomputes and commits the hash** on PRs that touch `go.mod`, so Dependabot's own branches self-heal. Needs a token that can push to the PR branch, and needs to not fight Dependabot's own force-pushes.
- **A check that fails with the right instruction** rather than a raw Nix hash mismatch - naming the file, the field, and the command that produces the new value. Cheapest option; does not remove the manual step but makes it a thirty-second one for anyone.
- **Drop the pin.** `vendorHash` exists because the Nix build vendors from the network; if the build can be made to use a checked-in vendor directory or a Go module proxy Nix already trusts, the hash stops existing. Larger change, and it should be argued rather than assumed.

## Verification

A Dependabot PR that bumps a Go dependency reaches a green `Nix build` job with no human editing `flake.nix`, or fails with a message that names exactly what to run.

atqamz:
## Summary

Three layered changes, so a Dependabot Go bump reaches a green `Nix build` with no human editing `flake.nix`:

- **`.github/dependabot.yaml`**: group all `gomod` updates into one PR. Two dependency PRs open at once meant the second one's correct `vendorHash` was only knowable after the first merged (atqamz/hand#446 needed two recomputes); grouping removes that case outright.
- **`.github/workflows/ci.yaml`**: `make vendorhash` (recomputes the correct hash via nixpkgs' fakeHash trick) plus a diagnostic step on `Nix build` failure that names the file/field/command instead of a raw Nix hash-mismatch trace.
- **`.github/workflows/vendorhash-selfheal.yaml`**: a credential-free self-heal path for real Dependabot PRs. The read-only-token `nix-build` job - the only one that ever fetches/builds dependency code the PR controls - hands off just a recomputed hash string as a build artifact once it's proven the fix builds. A separate, `workflow_run`-triggered, write-scoped workflow (which always runs the trusted default-branch copy of itself) never runs `go`, `nix`, or `make`; it independently verifies the branch's open PR was authored by `dependabot[bot]`, takes the branch/SHA only from the `workflow_run` event payload (never the artifact), re-validates the hash against a strict SRI shape, and refuses to push unless exactly one file changed by exactly one line. It then dispatches a fresh CI run, since a `GITHUB_TOKEN` push doesn't start one on its own.

Closes atqamz/hand#483

### Directions considered

- **Drop the `vendorHash` pin (checked-in vendor directory):** rejected with a measurement. `go mod vendor` against this repo's actual deps produced a 141MB, 2117-file tree - too costly to check into a small CLI repo permanently.
- **Self-healing CI:** initially rejected on a wrong premise (thought it needed a new PAT). It doesn't - re-verified against GitHub's current docs and a live throwaway probe: `workflow_run` gets a write-scoped `GITHUB_TOKEN` regardless of the upstream Dependabot `pull_request` run's restricted token. Built as described above once that held up, with the operator's two added conditions (validate the artifact's hash shape strictly; never trust the artifact for branch/SHA) incorporated.
- **Message-only diagnostic:** kept regardless, as the fallback signal and as what a human sees while working through any case the self-heal path doesn't cover.

Also found along the way, not acted on: this repo's branch ruleset has no required-status-checks rule at all today, so a red `Nix build` has never technically blocked merging here - the operator is raising that separately. And `github.actor`/`triggering_actor` on this repo's real Dependabot CI runs is the human approver, not `dependabot[bot]` (this repo can require approval to run a Dependabot PR's workflow) - both workflows key off the PR's actual author instead.

## Test plan

- [x] `make lint`
- [x] `make test`
- [x] `nix build .#default`
- [x] Reproduced the failure: bumped `golang.org/x/sys` and `modernc.org/sqlite` in `go.mod`/`go.sum` without touching `flake.nix`; confirmed `nix build .#default` fails.
- [x] Ran `make vendorhash` against that change; confirmed it computes a hash that makes `nix build .#default` succeed.
- [x] Set a deliberately wrong (but well-formed) `vendorHash`; confirmed `nix build .#default` still fails with the standard hash-mismatch error - a wrong hash is never green.
- [x] `actionlint` passes clean on both changed/added workflow files.
- [x] Simulated the self-heal apply/push logic against a real scratch git repo: happy path (clone/patch/commit/push) succeeds; a simulated Dependabot force-push race is correctly refused; an extra unexpected file change is correctly refused; the hash-shape regex accepts real SRI hashes and rejects malformed/injection-shaped strings.
- [x] Live CI run on this branch is fully green, including `Nix build (x86_64-linux)`, which exercises the new workflow YAML.
- [ ] Unproven until a real Dependabot PR arrives: whether `workflow_run` actually fires and gates correctly for a run whose upstream event came from a Dependabot-authored PR in this repo, and whether the dispatched CI re-run surfaces as checks on that PR. Both are backed by GitHub's current docs and (for the latter) a corroborating community report, but `workflow_dispatch`/`workflow_run` registration requires the workflow file to already be on the default branch, so this could not be exercised end-to-end from a branch.

### DIFF
--- a/.github/dependabot.yaml
+++ b/.github/dependabot.yaml
@@ -6,6 +6,9 @@ updates:
       interval: weekly
     commit-message:
       prefix: "chore(deps)"
+    groups:
+      go-modules:
+        patterns: ["*"]
   - package-ecosystem: github-actions
     directory: /
     schedule:
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -5,6 +5,10 @@ on:
     branches: [main]
   push:
     branches: [main]
+  # Lets the self-heal workflow (vendorhash-selfheal.yaml) re-run CI on a Dependabot
+  # branch after it pushes a corrected vendorHash: a GITHUB_TOKEN push does not start a
+  # new run on its own, so that workflow dispatches one explicitly.
+  workflow_dispatch:
 
 jobs:
   lint:
@@ -143,6 +147,35 @@ jobs:
           else
             echo "::warning::could not determine whether vendorHash is stale; see the 'Build flake package' failure above."
           fi
+      - name: Hand the fix to the privileged self-heal workflow
+        if: failure() && github.event.pull_request.user.login == 'dependabot[bot]'
+        # This job's token is read-only on a Dependabot pull_request run regardless of any
+        # permissions: block, and it is the only job that fetches or builds dependency code
+        # the PR controls. It hands off just a hash string; vendorhash-selfheal.yaml, which
+        # has write access, never runs go, nix, or make.
+        run: |
+          if git diff --quiet flake.nix; then
+            echo "vendorHash is already correct; nothing to hand off."
+            exit 0
+          fi
+          hash=$(grep -oP '(?<=vendorHash = ")[^"]+' flake.nix)
+          if ! [[ "$hash" =~ ^sha256-[A-Za-z0-9+/]{43}=$ ]]; then
+            echo "::error::computed vendorHash '$hash' is not a well-formed SRI hash; refusing to hand it off." >&2
+            exit 1
+          fi
+          if ! nix build .#default --no-link; then
+            echo "::error::flake.nix with the recomputed vendorHash still does not build; refusing to hand it off." >&2
+            exit 1
+          fi
+          mkdir -p /tmp/vendorhash-fix
+          printf '%s' "$hash" > /tmp/vendorhash-fix/vendorhash.txt
+      - uses: actions/upload-artifact@v7
+        if: failure() && github.event.pull_request.user.login == 'dependabot[bot]'
+        with:
+          name: vendorhash-fix
+          path: /tmp/vendorhash-fix/vendorhash.txt
+          if-no-files-found: ignore
+          retention-days: 1
       - name: Run built binary
         run: '"$(nix build .#default --no-link --print-out-paths)/bin/hand" --version'
 
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -127,6 +127,22 @@ jobs:
       - uses: DeterminateSystems/nix-installer-action@v22
       - name: Build flake package
         run: nix build .#default --no-link --print-out-paths
+      - name: Diagnose stale vendorHash
+        if: failure()
+        # go.mod/go.sum changes invalidate vendorHash but nothing keeps them in sync;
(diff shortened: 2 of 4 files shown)


atqamz:
Added the certain fix: grouped all gomod updates into one Dependabot pull request (\`.github/dependabot.yaml\`). That removes the compounding case atqamz/hand#446 hit outright - with one PR there is never a second one to invalidate the first's hash.

Still investigating \`workflow_run\` for a credential-free direction 1 before building anything on top; will report separately.

Issue #524, opened 2026-09-01
architecture(v19): resolve cutover writer exclusion vs read-only source constraint


atqamz:
## Resolution and authority

The source-writer-exclusion **architecture decision** is resolved. This issue does not certify the implementation or the 0.8.0 release.

Current immutable authority:

```text
anchor revision: 899e7a9aa2e383d3b3f91689fe5a897c5bab59ec
manifest: docs/architecture/v19-contracts/manifest-v2.md
#348 contract: docs/architecture/v19-contracts/348-cutover-archive-v2.md
#348 blob: 94c018d0bce8c2427d01b9af89e513cfb5f3ce96
```

The decision was mechanically explored by diagnostic PR #525, deliberately relocked by #527, and anchored by #528. The diagnostic remains historical evidence, not a production implementation to merge. The historical comment in this issue is not normative authority.

The accepted boundary remains:

```text
known SHARED snapshot
→ positively observed future-reader barrier, with no SHARED gap
→ OUR EXCLUSIVE transaction actually acquired
→ exact source, Fleet-local lock, and provider-quiescence revalidation
→ durable immutable original archive
→ explicit one-way frozen bridge
```

Neither a generic read transaction nor `BEGIN IMMEDIATE` establishes this boundary. PID/process-name scans, sleeps, pathname equality, one-time `MigrationLock` checks, and equal before/after hashes do not prove uninterrupted exclusion.

## Current implementation state

At `main` revision `47bc0e1e950dd59d3f816306e625e45da5b5dd4f`, the 5A1–5A3 primitives are no longer wholly unimplemented: #529–#532, #534, and #536 landed source eligibility, SQLite handoff, Fleet-local lock closure, durable-state classification, and Project/Treehouse/Herdr observation. Later #537–#551 added archive, freeze, import, publication, and recovery primitives. These are implementation progress, not evidence that ordinary startup/cutover is integrated or release-qualified; #348 owns that completion.

A source-level audit identifies a blocking implementation counterexample in `internal/store/cutovergate.go` and `internal/store/cutoverarchive.go`: raw `os.Open`/`Close` of the active database is used for hashing/copying while SQLite SHARED/barrier/EXCLUSIVE locks are supposed to remain held. With POSIX process-associated record locks, closing another descriptor of that same file drops the process's locks. SQLite explicitly documents this hazard: https://www.sqlite.org/howtocorrupt.html#_posix_advisory_locks_canceled_by_a_separate_thread_doing_close_

[...]

Issue #348, opened 2026-08-22
docs(architecture): lock v18→v19 cutover, WorkerInput non-fabrication, and legacy archival semantics


atqamz:
# v18→v19 cutover implementation tracker

This issue tracks implementation and verification. It does not amend repository-owned architecture. Technical completion is recorded here; issue state/lock/milestone are unchanged.

## Current authority

- Semantic manifest: `docs/architecture/v19-contracts/manifest-v2.md` at `899e7a9aa2e383d3b3f91689fe5a897c5bab59ec`; contract-set SHA-256 `03314c8f9dad1e0716062b1b06afa627b177295643581b1fbae3571d93d7e21c`.
- Cutover: `docs/architecture/v19-contracts/348-cutover-archive-v2.md`, Git blob `94c018d0bce8c2427d01b9af89e513cfb5f3ce96`.
- Relational authority: #344's immutable DDL/proof at `67ca4b35b773ef25ac9ff88cd1b16213153ed498`.
- Revision 1 at `6699745bb451e3daca072b4c958e5b79f31e4775`, `348-cutover-archive.md`, and `manifest.md` remain historical, not current authority. #524/#525/#527/#528 preserve the writer-exclusion redesign/probe provenance; a probe does not qualify the production implementation.

Revision 2 remains in force. The correction below is an implementation obligation, not a silently amended contract. Expected DDL impact: NONE. Any required normative change must instead land as a new versioned contract/manifest and corresponding proof before dependent implementation; preserve old artifacts.

## Verified implementation state

Static audit baseline: `main` at `47bc0e1e950dd59d3f816306e625e45da5b5dd4f` on 2026-09-11 UTC.

#517's exact schema-family boundary and subsequent cutover helpers are landed work, not a reason to rebuild the subsystem. In particular, `internal/store/cutovergate.go` implements the SHARED/barrier/EXCLUSIVE sequence, and `internal/store/cutoverarchive.go` implements digest-bound candidate staging. They do not yet establish the required continuous inter-process exclusion. No local reproduction, end-to-end cutover, or release/platform qualification is claimed by this audit.

## P0: raw source descriptor close can invalidate SQLite exclusion

Concrete source evidence at the baseline:

1. `acquireLegacyV18CutoverGateWithTimeout` calls `legacyV18CutoverFileSHA256` after establishing the SHARED snapshot, under the reader barrier, after candidate staging, and after acquiring EXCLUSIVE.
[...]

Issue #344, opened 2026-08-22
docs(architecture): regenerate and relock v19 with WorkerInput + WorkerWake before #339


atqamz:
# Canonical v19 relational authority and completed archive relock

This body identifies immutable DDL/proof authority and tracks its reviewed replacement. Comments are not normative architecture state. #339 owns implementation; #305 owns exact-candidate qualification.

**Current artifact: revision-2 #344 authority is landed and anchored by manifest-v3.** The prior 55-table artifact below remains immutable historical evidence and is not schema-compatible merely because both use `user_version=19`.

## Current immutable artifact

    repository: atqamz/hand
    permanent contract-set anchor: b2e11b5e14007eb46bdc1e820d4e085afd4caa57
    immutable content revision: 47e4e30092a729451d3eec2b3b592cd1ea395725
    manifest landing: 158c95e6d7b4c9ff98ecd1d525d0d355d48f0546
    manifest: docs/architecture/v19-contracts/manifest-v3.md
    manifest Git blob: acb7f77f256fa8aec0f46bee4f25882cfe05a7d9
    DDL path: docs/architecture/v19-v2.sql.gz
    Git blob SHA: 27080702c5efb9ec5366e3180ebe38d2fbee8c76
    stored gzip bytes: 11622
    stored gzip SHA-256: 465ebf6c1203a66558f2562660d67c3c40f44afc03982115177544c4bacb9f20
    reconstructed DDL bytes: 109227
    reconstructed DDL SHA-256: bd1081a14f3d5c52803aab40865dd489b3528c71d3ed801ea9560438ae5c8892
    schema fingerprint: 067f7ea28694f3dadf9cbc8e5b6e50fe42e77f7b36aaeb33fc121ef9772a1c8e
    schema-defined objects: 56 tables / 38 explicit indexes / 172 triggers
    PRAGMA user_version: 19
    proof runner: docs/architecture/v19-proof-v2.py.gz
    proof Git blob: 518c398d39e29586a370ec4199d9c63b976986d0
    relock manifest: docs/architecture/v19-relock-v2.md
    relock Git blob: 962db47fb45667c0c4e07ed3df7a2af69a7e4462

Authority: https://github.com/atqamz/hand/blob/158c95e6d7b4c9ff98ecd1d525d0d355d48f0546/docs/architecture/v19-contracts/manifest-v3.md

[...]

PR #525, opened 2026-09-01
test(store): probe v18 cutover writer exclusion and freeze

atqamz:
Diagnostic proof for #524 only. Draft; not an architecture change and not intended to merge.

Mechanical evidence established at exact head `5f985170f6ece5d60fc9f083bfef70b0d8bbc667`:

1. `PRAGMA query_only=1` before a SQLite write transaction is not viable: the production driver returns `SQLITE_READONLY`.
2. `BEGIN IMMEDIATE → query_only=1` blocks writers but is insufficient for cutover because RESERVED still admits new readers.
3. Exact v0.7.2 `hand project add/create` proves why reader admission matters: it can read authoritative DB state, then clone/legacy-Treehouse mutate before eventual DB registration.
4. A read-only SHARED snapshot can remain held while a second pinned connection requests `BEGIN EXCLUSIVE`; the waiting EXCLUSIVE establishes a reader barrier before the known SHARED reader is released.
5. The barrier is positively observed by requiring a fresh `mode=ro` reader to fail `SQLITE_BUSY`; that observation is not ownership proof by itself.
6. A rollback-journal writer raced after the SHARED snapshot can create journal state and be hard-killed without changing the committed main DB candidate; archive candidate bytes remain stable while SHARED+barrier coexist.
7. After releasing only the known SHARED reader, OUR already-requested `BEGIN EXCLUSIVE` completes; post-acquisition source state/digest must remain candidate-identical or production cutover must refuse.
8. `query_only=1` on the acquired EXCLUSIVE connection preserves the no-mutation inspection boundary; `query_only=0` is crossed only after immutable-original evidence is durable.
9. A one-way frozen bridge with exact original digest certificate, 21 unconditional guards over the seven exact v0.7.2 tables, and mechanism sentinel `user_version=22` blocks a prepared stale writer and makes a fresh v0.7.2 `Open` fail closed.
10. Released-v0.7.2-equivalent `Open` against exact canonical #344 v19 fails without changing DB bytes; the numeric `user_version=19` overlap does not require #344 relock.

CI run 1046 passed Contract, Lint, Nix x86_64, Ubuntu x64/ARM, Windows, macOS ARM/Intel, and E2E on this exact head.

Architecture consequence: revision-2 #348 must use `SHARED snapshot → observed new-reader barrier → OUR EXCLUSIVE → exact digest/state + Fleet-local lock/provider quiescence → immutable original archive → one-way freeze`. `BEGIN IMMEDIATE`, archive-before-gate, and PENDING observation alone are explicitly rejected as sufficient authority boundaries.

PR #526 consumes this evidence. Keep this diagnostic PR draft/non-authoritative and do not merge it into production code.

Canonical #344 DDL impact: **NONE**.

### DIFF
--- a/internal/store/cutover_hot_journal_test.go
+++ b/internal/store/cutover_hot_journal_test.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"crypto/sha256"
 	"database/sql"
+	"fmt"
 	"net/url"
 	"os"
 	"os/exec"
@@ -11,7 +12,7 @@ import (
 	"time"
 )
 
-func TestCutoverArchiveBeforeGateSurvivesHotJournalRecovery(t *testing.T) {
+func TestCutoverReadSnapshotAndWriterGateSurviveHotJournal(t *testing.T) {
 	home := t.TempDir()
 	db, err := Open(home)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


atqamz:
Closing this diagnostic-only PR as superseded, not merging it.

Its exact-head mechanical evidence (`5f985170f6ece5d60fc9f083bfef70b0d8bbc667`, CI #1046) was consumed by the revision-2 #348 contract that ultimately merged via #527 (`899e7a9aa2e383d3b3f91689fe5a897c5bab59ec`). The production cutover implementation has since landed the corresponding no-gap SHARED → observed reader barrier → OUR EXCLUSIVE gate and deterministic archive work in later slices.

Keeping the 629-line probe branch out of `main` preserves the original intent stated in this PR: diagnostic evidence only, non-authoritative, not production code.

Canonical #344 DDL impact: NONE.

PR #587, opened 2026-09-15
fix(project): resolve HTTPS rewrites before SSH normalization

### LINKED ISSUE #520: feat(project): register an https project URL by resolving it to ssh, instead of refusing it
## Problem

`hand project add https://github.com/owner/repo` cannot clone, because the pinned Git carries no
`git-remote-https`. Today the operator's only route is to retype the URL in ssh form.

That friction is not hypothetical: `gh` ships with `git_protocol = https` by default, so every URL an
operator copies out of `gh repo view`, `gh repo clone`, or the GitHub web UI is exactly the form
`hand project add` refuses.

## Not a re-litigation of atqamz/hand#440

atqamz/hand#440 established, with evidence, that the helper is **absent from the payload**, and that
`GIT_EXEC_PATH` therefore cannot fix it - "there is nothing to find at any exec-path". Confirmed again
here: the upstream release hand pins,
`darkvertex/static-git` 2.51.0 `git-binaries.linux-64bit.tar.gz`, contains exactly 8 entries and no
transport helper at all.

```
git  git-cvsserver  git-receive-pack  git-shell  git-upload-archive  git-upload-pack  gitk  scalar
```

sha256 `3a9e3a757ac1b5950dff03e4de74bab21d15e3bd3a389bceaf3ecedc72ab1f7c`, matching
`internal/toolchain/runtime.lock.json`.

This issue proposes something different in kind: not making https transport work, but making an
https-form **URL** work by resolving it to ssh before Git needs a helper it does not have. The
diagnosis work from #440 stays exactly as it is and remains the fallback for every host hand cannot
rewrite.

## Evidence that the rewrite path works

The pinned Git reads the global Git config, and a `url.<base>.insteadOf` rule makes an https URL clone
over ssh with no helper involved:

```
$ B=~/.secondhand/runtime/bundles/rd2343fe130ff5ba2-*/git/git

$ GIT_TERMINAL_PROMPT=0 "$B" ls-remote https://github.com/atqamz/mitemiru.git
git: 'remote-https' is not a git command. See 'git --help'.
fatal: remote helper 'https' aborted session

$ GIT_TERMINAL_PROMPT=0 "$B" -c 'url.git@github.com:.insteadOf=https://github.com/' \
    ls-remote https://github.com/atqamz/mitemiru.git
b22fa468a75340ead2b4fc185d8cba538c2d14f9	HEAD
b22fa468a75340ead2b4fc185d8cba538c2d14f9	refs/heads/main
```

## Two ways to spend that, and the tradeoff

**A. Normalize the URL at registration.** `hand project add` rewrites a recognized https URL to its ssh
form, stores the ssh URL, and logs one line saying it did so and why. Nothing persists in config, and
`internal/project/pr.go`'s deliberate choice to read `config --get remote.origin.url` rather than
`remote get-url` keeps working, because the stored value stays honest about the transport actually in
use.

**B. Inject the rule for hand's own runtime Git.** `GIT_CONFIG_COUNT` / `GIT_CONFIG_KEY_0` /
`GIT_CONFIG_VALUE_0`, the mechanism `cmd/smoke_test.go:22-24` already uses. Leaves the operator's
global config untouched, but the rule then has to be injected on **every** runtime Git invocation, not
just the clone: the stored remote stays https, so a worker's later fetch and push need it too. It also
puts hand in the position of resolving a URL to something other than what the registry records - the
exact ambiguity the `pr.go` comment goes out of its way to avoid.

A is narrower, has no ongoing obligation, and keeps one stored truth. Recommending A.

## Scope

- Recognized hosts only, github.com and gitlab.com to start; anything else keeps today's diagnosis
  verbatim.
- Skip the rewrite when `git_https_ready` is true, so a runtime that gains a helper later stops
  rewriting on its own.
- Never rewrite silently. The log line is the point: an operator who pasted https should be able to see
  why the registry holds ssh.

## Acceptance

- `hand project add https://github.com/owner/repo` registers the project, storing
  `git@github.com:owner/repo.git`, and prints why.
- `hand project add` against an unrecognized host still fails with the atqamz/hand#440 diagnosis, unchanged.
- With a Git that does carry `git-remote-https`, an https URL is stored as given.
- A worker's later fetch and push in that clone need no additional config.

## Operator workaround until this lands

`gh config set git_protocol ssh`, so copied URLs are already in the working form. A global
`url."git@github.com:".insteadOf` also works but is worse: it makes the
`credential."https://github.com".helper = !gh auth git-credential` config a dead letter by routing all
GitHub traffic over ssh. Scoping such a rule with `includeIf gitdir:` does not help either, since the
failure happens at clone time, before the gitdir exists.

atqamz:
## Merged after manual review

Squash-merged as `e37693bdd5633033602385fd9ebda3c7253d83e6` on 2026-09-21.
Reviewed head: `23ccf3c8dd05670c8282895cec8a5e44e11e1c13`.
Manual review: `5267409469`, under explicit operator authorization to review and squash merge acceptable work without waiting for bot capacity.

## Scope

Use one managed Git Runtime for helper inspection, no-network `ls-remote --get-url`, and the single clone. Inspect reported exec path, managed GitBin and inherited PATH. Preserve original HTTPS on inspection errors. Normalize only positively helper-free supported GitHub/GitLab locators with no effective rewrite. Preserve credentials, trust/config, one destination reservation and no stderr-authorized retry.

Preserve meaningful whitespace in both Git exec paths and effective URL rewrites. Remove only the exec-path output terminator. For URL normalization require the byte-exact unchanged locator plus LF, rather than trimming response bytes.

Fixes atqamz/hand#520

## Review correction and verification

Manual review found that TrimSpace on `ls-remote --get-url` hid real insteadOf rewrites ending in space, tab, carriage return or NBSP and could authorize unintended SSH normalization. The fix has four real-Git regressions and an unchanged-locator positive control.

Workflow `35606523311` ran actual repository code with pinned Go 1.26.5 on Linux, integrated main `86ffe740`, reproduced all four valid failures RED, then passed the five-case regression 20 times under race. Full tagged cmd/toolchain race suites, vet, contract, formatting and diff checks passed before publishing the reviewed head. An initial fixture with newline in the Git config key was invalid and removed; it was not counted as a valid regression.

Standard exact-head CI `35607012233` subsequently completed successfully: all five platform Test/Build jobs, four E2E jobs, Nix, lint, contract and five mutation gates passed. The optional expensive mutation job was skipped. The former in-progress status is superseded by this completed result.

The original exec-path whitespace fix also retained its five real-helper controls. Later local verification used the complete repository and Go 1.26.5; the new rewrite regression passed 20 race repetitions as part of the integrated candidate checks.

The cross-PR fixture interaction with #590's deterministic generation paths was fixed and tested on #590, now also merged. No replacement PR, dependency, schema or permanent workflow change was introduced here. Native provider/runtime and 0.8.0 release qualification remain separate.


atqamz (commented):
Manual review completed on 23ccf3c8dd05670c8282895cec8a5e44e11e1c13 under the operator's explicit manual-review/squash-merge instruction.

Reviewed the transport/helper/clone flow and tests. Found and fixed one remaining blocker: TrimSpace on `git ls-remote --get-url` could hide a real insteadOf rewrite containing trailing space, tab, CR or NBSP and authorize an unintended HTTPS-to-SSH route change. The candidate now requires byte-exact unchanged locator plus its LF terminator; every other response retains the original locator. It preserves the single managed Runtime, original credentials/config route, positive helper absence, one clone/reservation, and no stderr-authorized retry.

Actual repository verification, not an extraction scaffold: workflow 35606523311 used pinned Go 1.26.5 on Linux, integrated main 86ffe7406ad7a3dcff5e62f6c52e122fe1d53c32, proved all four valid real-Git rewrite regressions RED, then passed the five-case regression 20 times under race, full tagged race tests for ./cmd and ./internal/toolchain, vet and make contract. A first test attempt included newline in a Git config key; Git rejected the fixture itself, so that invalid case was removed rather than counted as a regression. The workflow published this exact candidate only after those checks passed.

No remaining source-review blocker in this PR. Standard full-platform CI 35607012233 is still in progress; the already successful focused verification does not claim a completed platform matrix. Merge only the reviewed candidate after its remaining CI finishes. No bot approval, protection override, new PR, schema change or release qualification is implied.

PR #592, opened 2026-09-15
fix(store): pin legacy cutover source descriptor

atqamz:
## Summary

- pin the exact legacy SQLite source descriptor before locking and reuse bounded readers for every protected hash and copy
- retain source SQLite and ambiguous artifact descriptors until ordered SQL teardown, with fail-closed identity, link, size, alias, and replacement checks
- prove old POSIX fresh-open behavior and corrected initial/barrier/archive/EXCLUSIVE/freeze exclusion with deterministic independent-process handshakes

Refs atqamz/hand#348

## Test plan

- `go test ./internal/store -run Cutover -count=1 -timeout=5m`
- independent-process suite, 20 repetitions
- affected independent-process/error-path suite under `-race`, 5 repetitions
- `go test ./internal/store -count=1 -timeout=30m`
- `make contract`
- `make lint`
- CGO-disabled cross-compiles: Linux arm64, Windows amd64, macOS amd64/arm64


atqamz (commented):
Manual review completed on exact head 33b4db7f9d1986368be78af021d7921ae2719c71, including the full fifteen-file diff and the complete gate acquisition/teardown flow. No blocking finding in this source-descriptor correction.

The source descriptor and auxiliary SQL connections are acquired before the SHARED snapshot and retained through the EXCLUSIVE handoff. Hash/copy operations reuse the pinned descriptor with bounded SectionReaders and identity/link/size checks. Artifact opens reject source aliases; descriptors whose identity is uncertain or aliases the source are retained rather than closed while SQLite exclusion depends on them. Error and committed-freeze teardown release the SQL resources before the pinned descriptors. Manifest/archive paths consistently use the source-aware helpers while that source is live.

The independent-process tests cover the old fresh-open/close counterexample, writer exclusion at actual acquisition phases, explicit gate release, frozen-writer refusal, and archive/manifest alias failures. Rechecked exact-head CI 35055704572: completed/success. Existing code/security review comments also name this exact head; no unresolved review threads were returned. No new local test execution is claimed.

Ready for squash merge under the operator's explicit authorization. This fixes a cutover mechanism defect; it does not assert completion of #348, canonical runtime integration, or release qualification.

